PR #15, opened 2014-10-23
Register user with form's save method, whenever possible

laginha:
Hey,

In a nutshell, as I added to the `CHANGELOG`:

* Feature: Added settings' options that allows to exlude the default auth urls (`INCLUDE_AUTH_URLS`) and register url (`INCLUDE_REGISTER_URL`).

* Feature: Added support to dynamically import any chosen registration form using the settings option `REGISTRATION_FORM`.

* Enhancement: Make `RegistrationForm` a subclass of Django's `UserCreationForm`

* Enhancement: Use registration form `save` method to create user instance, whenever form is a subclass of Django's `ModelForm`.

Basically I wanted to use your useful app in a project I'm working on, but with a few twists:

* I didn't want to use the `register/` url, because I have two distinct profiles each one with a different register form and view.

* I wanted to use a registration form of my own, which is a subclass of `ModelForm`, with logic in the `save` method, associated with a custom user with no username field.

I've run the tests and everything is working. I've also updated the docs, version number and changelog.

Hope you approve my work.
If you have any doubt, please contact me and I'll try to explain things better.

DLM

### DIFF
--- a/registration/auth_urls.py
+++ b/registration/auth_urls.py
@@ -51,10 +51,6 @@ urlpatterns = patterns('',
                            auth_views.password_reset,
                            {'post_reset_redirect': reverse_lazy('auth_password_reset_done')},
                            name='auth_password_reset'),
-                       url(r'^password/reset/confirm/(?P<uidb64>[0-9A-Za-z_\-]+)/(?P<token>.+)/$',
-                           auth_views.password_reset_confirm,
-                           {'post_reset_redirect': reverse_lazy('auth_password_reset_complete')},
-                           name='auth_password_reset_confirm'),
                        url(r'^password/reset/complete/$',
                            auth_views.password_reset_complete,
                            name='auth_password_reset_complete'),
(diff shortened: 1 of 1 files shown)


macropin:
Sorry it's taken me a while to get back to reviewing this. It's a big commit so I'd like to look at it carefully. Are the urlconf defaults aligned to the current behaviour?

laginha:
I did not understand what you mean...

The urlconf defaults behaves as it did before, by default. The new settings I included  (`INCLUDE_REGISTER_URL` and `INCLUDE_AUTH_URLS `) are optional.
Every other changes I made are backward compatible, I believe.

macropin:
Can you please also take a look at _default/views.py_. That is not using form.save(). Thanks.

laginha:
It is [if form has a `save` method](https://github.com/macropin/django-registration/blob/master/registration/backends/default/views.py#L91). Did I understand you wrong?

macropin:
Nevermind. I think I was lacking caffeine yesterday. :)

PR #20, opened 2014-12-09
specify Python versions

chhantyal:
While we are at it, let's this tool https://github.com/brettcannon/caniusepython3 know about Python versions

btw, thanks a ton for maintaining up to date fork. I just replaced legacy package with this in all my projects :)


macropin:
Thanks.

PR #23, opened 2015-01-26
Make success_url an attribute on registration views

stianjensen:
This makes it possible to override the success url without subclassing
the registration view, simply by passing a success_url in from the
URLConf


macropin:
I'm puzzled as to why it was ever implemented that way. Thanks

PR #24, opened 2015-01-28
use get_username() from AbstractBaseUser

pahaz:
https://docs.djangoproject.com/en/1.7/topics/auth/customizing/#django.contrib.auth.models.AbstractBaseUser.get_username


macropin:
Can you please fix this for Django==1.4.15

macropin:
Thanks for this

PR #35, opened 2015-03-19
Flatten variable use in blocktrans.

### LINKED ISSUE #29: Templates blocktrans refactoring
Blocktrans in templates
https://github.com/macropin/django-registration/tree/master/registration/templates/registration
does not support dict, list, etc. So this record

```
{{ site.name }}
```

in blocktrans is wrong and does not work.

it must be something like that

```
{% blocktrans with site_name=site.name %}
{{ site_name }} registration
{% endblocktrans %}
```

dracos:
Fixes #29.
The translations will need updating.

### DIFF
--- a/registration/templates/registration/activation_email.html
+++ b/registration/templates/registration/activation_email.html
@@ -9,9 +9,9 @@
 
 <body>
 <p>
-    {% blocktrans %}
+    {% blocktrans with site_name=site.name %}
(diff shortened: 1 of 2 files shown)


macropin:
I don't think site.name is something we would want to translate.

macropin:
Thanks

Issue #11, opened 2014-09-17
Templates are not installed


macropin:
I believe that we should ship and install a base set of templates that at least provide a starting point.

Currently the templates included here are not installed by default. Modification to `setup.py` is required.

Previously the following was given as the reason for not shipping templates.

**Does |project| come with any sample templates I can use right away?**
    Yes, for two reasons:

    1. Providing default templates with an application is generally
       hard to impossible, because different sites can have such
       wildly different design and template structure. Any attempt to
       provide templates which would work with all the possibilities
       would probably end up working with none of them.

    2. A number of things in |project| depend on the
       specific :ref:`registration backend <backend-api>` you use,
       including the variables which end up in template
       contexts. Since |project| has no way of knowing in
       advance what backend you're going to be using, it also has no
       way of knowing what your templates will need to look like.
    
    Fortunately, however, |project| has good documentation
    which explains what context variables will be available to
    templates, and so it should be easy for anyone who knows Django's
    template system to create templates which integrate with their own
    site.

derek-adair:
I think this should ABSO-FREAKING-LUTELY come with a basic set of templates.  I was pretty shocked it didn't come with a basic working template which would at least represent the most common use cases.

There should at *least* be a bare bones working template so when someone installs the plugin they dont get a freaking error.  Ideally, *all* of the different use cases would have a template and users can just... reference the template!

It just seems like lazy excuses not to include some working examples here.

derek-adair:
I'm willing to submit a PR with a bare bones template as well.

derek-adair:
Sorry i dropped the ball here, something about a 2 week vacation and getting married i seemed to have lost this issue in my todo-lists :)  Life is back to normal now so I am available... for reals this time.

PR #38, opened 2015-03-28
include templates

douwevandermeij:
I added the templates of the package to manifest.in to be installed using setup.py and added a newline to the end of the file (pep8).


macropin:
Ok lets install them. Hopefully this doesn't break too many installs due to template issues. Refs #11

PR #42, opened 2015-04-17
Fix warning on using ForeignKey vs OneToOneField

dashmug:
Using `django-registration-redux` reveals the following warning:

```
WARNINGS:
registration.RegistrationProfile.user: (fields.W342) Setting unique=True on a ForeignKey has the same effect as using a OneToOneField.
	HINT: ForeignKey(unique=True) is usually better served by a OneToOneField.
```

This pull request simply fixes that issue by replacing the user field with a OneToOneField.

### DIFF
--- a/registration/models.py
+++ b/registration/models.py
@@ -186,7 +186,7 @@ class RegistrationProfile(models.Model):
     """
     ACTIVATED = "ALREADY_ACTIVATED"
 
-    user = models.ForeignKey(UserModelString(), unique=True, verbose_name=_('user'))
+    user = models.OneToOneField(UserModelString(), verbose_name=_('user'))
     activation_key = models.CharField(_('activation key'), max_length=40)
 
     objects = RegistrationManager()


jose-lpa:
+1 Easy one, should be merged to fix that `fields.W342` message.

dashmug:
Yes, and I think there wouldn't be any harmful side-effect.

macropin:
Thanks. Looks good.

nikolas:
Shouldn't there be a migration associated with this change?

jose-lpa:
@nikolas Not really necessary, since `OneToOneField` has a similar DB schema under the hood than a `ForeignKey(unique=True)`, so no DB operations will be applied with this specific change.

nikolas:
Hmm .. when I switched a field in one of my local apps from a `ForeignKey(unique=True)` to a `OneToOneField`, `makemigrations` generated a migration. Maybe there was something else going on ..

nikolas:
Actually, applying this change manually to django-registration in one of my apps generated this migration:

`worth2/migrations/registration/0002_auto_20150423_1251.py`
```py
# -*- coding: utf-8 -*-
from __future__ import unicode_literals

from django.db import models, migrations
from django.conf import settings


class Migration(migrations.Migration):

    dependencies = [
        ('registration', '0001_initial'),
    ]

    operations = [
        migrations.AlterField(
            model_name='registrationprofile',
            name='user',
            field=models.OneToOneField(verbose_name='user', to=settings.AUTH_USER_MODEL),
        ),
    ]
```

I'd think this should be included with django-registration, but I'm not sure.

jose-lpa:
But that is your own app in your own project, right? Not a 3rd party package like this one ;)

Django registration never had migrations. In any case, if you want to do it, 2 migrations should be added: the initial one with the previous status before this code was merged and the one which applies the code changes in the model.

However, this case doesn't make any changes in the DB schema at all, because roughly a `OneToOneField` implements the same as a `ForeignKey(unique=True)`, with some extra [https://docs.djangoproject.com/en/1.8/ref/models/fields/#onetoonefield](utilities at Python level), not at DB level. When syncing a model with a `ForeignKey(unique=True)` or a model with a `OneToOneField()`, the resulting DB table rows will be exactly the same.

I'm actually not sure anyway why Django doesn't just pass this specific change when doing `makemigrations`, since the new migration doesn't add anything relevant. In fact, one can just fake it and the system won't notice at all, except for those parts in the code where Python is expecting a `OneToOneField`, of course.

nikolas:
The migration only gets put in my application because I had to put this in my `settings.py` in order for django-registration to be compatible with Django 1.8:

```py
REGISTRATION_APPLICATION_MODEL = 'registration.Application'

MIGRATION_MODULES = {
    'registration': 'worth2.migrations.registration',                                                                                                                       
}
```

Once #25 is merged, it's necessary for this migration to be part of django-registration, because `makemigrations` triggers a migration for the field, even if there's no underlying changes in the db.

jose-lpa:
Oh, I see that #25 now :P 
So, true, if migrations are becoming mandatory for all apps, then yeah, this needs that migration. My whole argument was that if there are no migrations at all for this independent app, for this change won't worth creating migrations, because it is not a DB level change at all, so data integrity won't notice.

However, if there is a plan to have migrations for Django registration, it is necessary, yes. Even better: the #25 must be updated and just make an initial migration from this code, instead of from the old one.

nikolas:
Yeah good point, #25 should be updated to reflect current master. That would be simplest.

PR #47, opened 2015-05-01
Run tests against Django 1.8

AMeng:
The error message for invalid usernames was [changed in Django 1.8](https://github.com/django/django/blob/1.8/django/contrib/auth/models.py#L390).


macropin:
Thanks

PR #49, opened 2015-05-05
Filter user passwords from appearing in error reports

skolsuper:
Filter user passwords from appearing in error reports from RegistrationView as recommended in django docs [here](https://docs.djangoproject.com/en/1.7/howto/error-reporting/#django.views.decorators.debug.sensitive_post_parameters)


macropin:
Thanks

PR #52, opened 2015-05-10
Document the need for base.html

### LINKED ISSUE #34: base.html
demo form registration_base_.html  extends base.html .. which does not exist ..  Is that an oversight or something expected to be written by user ?

AMeng:
Fixes #34


macropin:
Thanks

PR #53, opened 2015-05-10
Correct docs around successful registration redirect

### LINKED ISSUE #12: “simple” (one-step) backend redirects not redirecting
Turning on the one-step backend as described [here](https://django-registration-redux.readthedocs.org/en/latest/simple-backend.html) does not redirect the user to get_absolute_url() after registration, instead to the usual /accounts/register/complete/.

AMeng:
Fixes #12


macropin:
Thanks for your fixes.

PR #55, opened 2015-05-10
Test against Python 3.4 and latest Django releases

AMeng:
The `setup.py` file indicates that Python 3.4 is supported. I've added that to the travis builds. Also bumped to the latest Django minor versions.


macropin:
Thanks

Issue #50, opened 2015-05-06
Can't override email template as Django returns original as first match


riklaunim:
After upgrading from old django-registration to redux version my email template rendering blow up. 

When I use django:
   loader.select_template(['registration/activation_email.txt'])
I get the django-registration original template (while my overriden template is in my_app/templates/registration/activation_email.txt). The my_app/templates is discovered by AppDirectory template loader, it's on the django.template.loaders.app_directories.app_template_dirs folders list, but after "registration" app templates folder (as it's after "registration" on the INSTALLED_APPS lists).

select_template calls get_template which returns first match. For some reason it worked before, but now (Django 1.7.7 and @master django-registration-redux) it does not. I can hack it by providing dirs value (either filtered or reversed):

        from django.template.loaders.app_directories import app_template_dirs
        dirs = sorted(app_template_dirs, reverse=True)
        template = loader.select_template(['registration/activation_email.txt'], dirs=dirs)

But it would be good to know why and what is going on. It should "just work".

AMeng:
I'm able to reproduce this. Appears to have been introduced in https://github.com/macropin/django-registration/commit/12aba225843023cc7355406e7c984eec733f4a2d when templates were added to the application. Django version should not matter here.

**To reproduce this:** Override `registration/activation_email.txt` but do *not* override `registration/activation_email.html`. The default `.html` template will be used in the emails.

**Cause:** The issue is that the code first looks for the `.txt` version of the template, and then checks for the `.html` version. It finds your `.txt` version just fine, but then finds the newly included `.html` template, and uses that instead.

Before https://github.com/macropin/django-registration/commit/12aba225843023cc7355406e7c984eec733f4a2d, no templates were included, so if you never created a `.html` version, your `.txt` version would be used.

AMeng:
@riklaunim this was addressed in https://github.com/macropin/django-registration/pull/57.

The solution is to either override the `registration/activation_email.html` template or add the following to settings.py:

```python
REGISTRATION_EMAIL_HTML = False
```

riklaunim:
I'll check this out.

riklaunim:
I have both .txt and .html templates

riklaunim:
This doesn't solve the problem. I've override both templates and more and still the .txt template is fetched in the original form (/env/lib/python3.4/site-packages/registration/templates/registration/activation_email.txt) which results in

[...]

PR #57, opened 2015-05-12
Add setting for REGISTRATION_EMAIL_HTML that defaults to True.

AMeng:
This addresses the issue raised in https://github.com/macropin/django-registration/issues/50.

Essentially, we just need to document the fact that both templates now need to be overridden. Alternatively, users can add a setting to disable the HTML template.

There is no easy way to determine weather or not a user has overridden one template but not another. It might be possible to inspect the paths of the loaded templates, but that seems like a hack.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -6,21 +6,32 @@ django-registration-redux changelog
 Version 1.2, 11 May, 2015
 -------------------------------
 
-* BACKWARDS-INCOMPATIBLE CHANGE: An initial migration for Django > 1.7 has been provided.
-  South users will need to configure a null migration with (`SOUTH_MIGRATION_MODULES`).
+* BACKWARDS-INCOMPATIBLE CHANGE: When overriding the
+  `registration/activation_email.txt` template, you must also now override the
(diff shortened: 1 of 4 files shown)


macropin:
Thanks

martey:
Since this setting defaults to True, isn't this a "backwards incompatible change" from the original django-registration (which did not use HTML email templates)?

I think it makes more sense to set this to False and document that it can be changed to enable HTML templates, rather than hoping that upgrading users read the CHANGELOG.

AMeng:
@martey, If we set this to `False` by default, and users have their own custom `.html` template, it will completely ignore that template and use their `.txt` version.

martey:
@AMeng, if we set this to `True` by default and users do not have a custom `.html` template (because they are upgrading from django-registration 1.0, like @riklaunim in #50), won't this ignore their custom `.txt` template?

At the very least, this setting (and the fact that the default HTML templates will supersede text templates) should be added to the "Backwards-incompatible changes" section of the upgrade guide in the documentation.

Issue #27, opened 2015-02-05
Can't specify RegistrationForm (form_class) to be used by the default backend


atleta:
It seems that there is no way to set the form_class for backends.default.RegsitrationView . The documentation says something about providing a keyword argument to 'registration()' (sic), but it seems to have been left there from an earlier version. At least I can't interpret it.

Also, I have looked through the code and can't see any sign of using anything but using the wired in form_class (specified in registration.views.RegistrationView). I would make it a setting in settings. If you are OK with it, I can create a pull request.

(Note, that I'm trying to upgrade from 0.8.0 where this could be handled through providing my own backend class.)

macropin:
I suspect we need the docs updated here. I'll mark this as a bug

wfranzini:
I've the same problem, but it seems that feature (REGISTRATION_FORM) is not available in v1.1.
Maybe the documentation refers to the development version.

martey:
I think I ran into the same problem while upgrading from django-registration. I had previously subclassed RegistrationView with a custom `form_class`; django-registration-redux did not work until I added a `REGISTRATION_FORM` setting.

This took more time than I would have liked to fix, because the documentation (immediately below the information about `REGISTRATION_FORM`) claims that overriding `form_class` works and that there are no backwards incompatible changes when upgrading from django-registration.

macropin:
I believe this is resolved by the release 1.2, about to go out.

martey:
I tried 1.2c0, but this isn't resolved.

Before 4d9c72e000a1a0a22b04f7b6f12b79634eadbeb1, you could use a custom registration form by subclassing RegistrationView and using a custom form_class. After that commit, you must use `REGISTRATION_FORM`.

I don't see using a setting as superior to subclassing (especially since this breaks backwards compatibility with the original django-registration), but maybe @laginha can explain.

laginha:
Subclassing `RegistrationView` and overriding `form_class` should work. Can you provide details of the error?

martey:
If I am using a custom user Model (`EmailUser`) that does not have the username field and do not set the `REGISTRATION_FORM` setting, I get the following error:
````python
FieldError: Unknown field(s) (username) specified for EmailUser

  File "/home/vagrant/.virtualenv/local/lib/python2.7/site-packages/registration/views.py", line 22, in <module>
    REGISTRATION_FORM = import_string( REGISTRATION_FORM_PATH )
[...]

PR #58, opened 2015-05-12
fix issue #27

laginha:
Use `USERNAME_FIELD` attribute to define the `RegistrationForm` meta fields (useful for custom user models)

### DIFF
--- a/registration/forms.py
+++ b/registration/forms.py
@@ -16,6 +16,7 @@ from django.contrib.auth.forms import UserCreationForm
 
 from registration.users import UserModel
 
+User = UserModel()
 
 class RegistrationForm(UserCreationForm):
     """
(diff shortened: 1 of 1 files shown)


AMeng:
Would be good to add a test for this.

macropin:
Looks good, but as @AMeng said a test using a custom User model would be a good idea. Can you please add one.

AMeng:
Added a test here: https://github.com/macropin/django-registration/pull/61

AMeng:
I added a couple of tests for this in https://github.com/macropin/django-registration/pull/61. As I mentioned there this fix really just stops the form import from breaking in versions of Django < 1.8. After that it does actually let you subclass the form successfully. I guess that should be documented.

PR #59, opened 2015-05-12
Reorganize backwards incompatible changes in "upgrade guide".

martey:
Reorganize backwards incompatible changes by affected version. Add information about two previously undocumented changes.


macropin:
Great idea. Especially since we've got a few more coming.

PR #61, opened 2015-05-15
Support custom user models with the RegistrationForm. Add tests.

### LINKED ISSUE #27: Can't specify RegistrationForm (form_class) to be used by the default backend

AMeng:
Fixes #27. I also fixed the same issue with the `ModelAdmin`.

### DIFF
--- a/registration/admin.py
+++ b/registration/admin.py
@@ -4,19 +4,20 @@ from django.contrib.sites.models import Site
 from django.utils.translation import ugettext_lazy as _
 
 from registration.models import RegistrationProfile
+from registration.users import UsernameField
 
 
 class RegistrationAdmin(admin.ModelAdmin):
     actions = ['activate_users', 'resend_activation_email']
     list_display = ('user', 'activation_key_expired')
     raw_id_fields = ['user']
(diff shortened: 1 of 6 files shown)


AMeng:
This doesn't seem to work for versions of Django before 1.8. According to the [Django docs](https://docs.djangoproject.com/en/1.8/topics/auth/customizing/#custom-users-and-the-built-in-auth-forms):

> UserCreationForm
> Depends on the User model. Must be re-written for any custom user model.

Because `RegistrationForm` subclasses `UserCreationForm`, it cannot be used for custom user models. At least not in versions of Django before 1.8.

The correct fix might be updating the docs and explaining this.

AMeng:
Squashed the commits into one.

PR #69, opened 2015-06-03
Solved "RemovedInDjango19Warning" when `sites` is not installed

philiptzou:
This patch:

1. solved `RemovedInDjango19Warning` when `django.contrib.sites` is not installed
2. used relative import to import internal modules
3. did some PEP8 fix


macropin:
Thanks.

PR #72, opened 2015-06-15
Distinguish between activation failure and attempted reactivation. Fix #70

### LINKED ISSUE #70: Activation failed - email clients likely pre-fetching the activation URL?
Has anyone else run into email clients, specifically web based ones, pre-fetching the activation URL? This causes the User's email to be verified and therefore activated. When the User subsequently requests the activation URL, they are presented with a failure message, but their account has actually already been activated.

Given the design of setting the activation key to `ALREADY_ACTIVATED`, and passing the requested activation key back to the activation complete view upon failure, I don't see a straightforward way to distinguish between failures due to the User already being activated and every other failure scenario.

dicato:
Add a boolean field , `activated` and corresponding migration
to `RegistrationProfile`. Use this field when the user tries
to activate their account to distinguish between:

- account activation failure due to non-existant activation key
- account activation failure due to an expired activation key
- account activation failure due to the user trying to activate
  more than once

By distinguising between these scenarios, it is possible to
provide the user with better information during account activation.
Specifically, if the user is active and tries to reactivate,
clearly indicate their account is active and able to login.

Update documentation and unit tests to account for additional
model field and activation logic.

Add both a schema and data migration. Data migration checks
if `activation_key` is the `ACTIVATED` string and conditionally
sets the new `activated` field.

---

I dug into this after having multiple users get confused by activation failure messages. See #70 for context.

### DIFF
--- a/docs/default-backend.rst
+++ b/docs/default-backend.rst
@@ -33,15 +33,15 @@ This backend makes use of the following settings:
     not supplied.
 
 ``INCLUDE_AUTH_URLS``
-    A boolean (either ``True`` or ``False``) indicating whether auth urls 
-    (mapped to ``django.contrib.auth.views``) should be included in the 
+    A boolean (either ``True`` or ``False``) indicating whether auth urls
+    (mapped to ``django.contrib.auth.views``) should be included in the
     ``urlpatterns`` of the application backend.
-    
+
 ``INCLUDE_REGISTER_URL``
-    A boolean (either ``True`` or ``False``) indicating whether the view 
-    for registering accounts should be included in the ``urlpatterns`` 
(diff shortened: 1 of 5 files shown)


dicato:
@macropin if there is anything I can do help with this, please don't hesitate to ask!

dicato:
I pushed an additional test recently that explicit tests trying to re-activate an account manually deactivated by a Django Admin.

@macropin is there anything I can do to help get this reviewed and hopefully merged? I've been running this branch in production since June 15. :smile:

macropin:
Thanks.

dicato:
Thank you  :+1:

PR #78, opened 2015-07-12
Added south initial migrations

mauler:
Despite you configure South to look into
registration.south_migrations modules no table will be created
if there is no 0001_initial

### DIFF
--- a/registration/south_migrations/0001_initial.py
+++ b/registration/south_migrations/0001_initial.py
@@ -1,0 +1,191 @@
+# -*- coding: utf-8 -*-
+from south.utils import datetime_utils as datetime
+from south.db import db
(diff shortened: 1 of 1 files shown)


mick-t:
This merge will fix my [issue](https://github.com/macropin/django-registration/issues/87). When will it be merged as I don't see this in the [master](https://github.com/macropin/django-registration/tree/master/registration/south_migrations) branch?

macropin:
@mick-t Thanks for confirming this . Merged.

mick-t:
I tired to use these migrations and they fail for me. Looks like mauler included some custom models in the migration that I didn't notice when I initially glanced at his merge request. See my merge request.

PR #84, opened 2015-08-05
mention blocks needed in base template

kevinmickey:
This may be obvious to some, but it would help to identify what blocks are needed in the base.html template.


macropin:
Thanks

PR #90, opened 2015-08-19
add 'remember me' functionality on the login form

Ilias95:
Add a 'remember me' checkbox on the login form. If it is not checked
during login, the user's session cookie will expire when user's browser
is closed. Otherwise, it will follow the default expiry policy.

### DIFF
--- a/registration/auth_urls.py
+++ b/registration/auth_urls.py
@@ -32,7 +32,7 @@ from django.contrib.auth import views as auth_views
 
 
 def login_user(request, template_name):
-    if not 'remember_me' in request.POST:
+    if request.method == 'POST' and not 'remember_me' in request.POST:
         request.session.set_expiry(0)
     return auth_views.login(request, template_name)
 


dicato:
I'm not sure this belongs directly in the this project as it deals directly with authentication, not user registration.

dicato:
Does anyone else have opinions on this?

joshblum:
This case is covered in django-allauth, https://github.com/pennersr/django-allauth/blob/cf45cc605a3d2b49efb1309960df7ee6059175cd/allauth/account/forms.py#L59, if the PR can be completed (with tests and the feature should be optional probably) I don't see a problem accepting it.

registration/auth_urls.py
```diff
@@ -31,9 +31,15 @@
 from django.contrib.auth import views as auth_views
 
 
+def login_user(request, template_name):
+    if request.method == 'POST' and not 'remember_me' in request.POST:
+        request.session.set_expiry(0)
```
joshblum: does the django auth view automatically handle if this parameter is set? Or should a line setting the cookie be placed? (i.e. https://github.com/pennersr/django-allauth/blob/cf45cc605a3d2b49efb1309960df7ee6059175cd/allauth/account/forms.py#L159)

Ilias95:
@joshblum
Regarding the line note. If I am not wrong, SESSION_COOKIE_AGE must be the default expiry policy anyway. Check: https://docs.djangoproject.com/en/1.9/topics/http/sessions/#django.contrib.sessions.backends.base.SessionBase.get_expiry_age . But if we set a different expiration policy somewhere else, this policy will be followed. So maybe we need to declare this explicitly? I am not sure what would be the desirable action in this case.

Regarding the tests, I'm sorry guys but I'm not a django pro and I haven't written any django code since pretty much I opened this PR. So I don't really remember how to write these kind of tests.

joshblum:
@Ilias95 no problem! You can find examples of tests here: https://github.com/macropin/django-registration/blob/master/registration/tests/simple_backend.py, I would recommend using this file to create your own test that sends the remember me parameter in the request

registration/auth_urls.py
```diff
@@ -31,9 +31,15 @@
 from django.contrib.auth import views as auth_views
 
 
+def login_user(request, template_name):
```
joshblum: I think it would be cleaner to move the `remember_me` functionality into a form, as django-allauth does. see https://github.com/pennersr/django-allauth/blob/cf45cc605a3d2b49efb1309960df7ee6059175cd/allauth/account/forms.py#L151

registration/templates/registration/login.html
```diff
@@ -7,6 +7,7 @@
 <form method="post" action="">
     {% csrf_token %}
     {{ form.as_p }}
+    <p><label><input type="checkbox" name="remember_me" />{% trans "Remember Me" %}</label></p>
```
joshblum: if we include the remember_me in the form, we don't need to manually add the line here

Ilias95:
Ok, I will try to implement the suggested changes and the tests in the next days. Git question: is there any way to sum up all the different "uneeded" commits in my branch to a final complete commit? Or this can be done after merging?

joshblum:
Checkout http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git!

joshblum:
@Ilias95 any update on this PR?

Ilias95:
@joshblum  I'm sorry, I haven't been using django for a big while so I really don't remember its ecosystem. Besides, I really don't have the time to look it up at the moment, even though it may be trivial.
Feel free to delete it or "fix" it.
Thanks.

joshblum:
@Ilias95 will leave this open in case someone else can take it over

joshblum:
Going to close this for now, we can readdress if there is a need!

Issue #87, opened 2015-08-10
Syndb and south migration fails with Django 1.5.11: You have no migrations for the 'registration' app. You might want some.


mick-t:
Hi,

I'm using Django 1.5.11 and I've installed django-registration-redux 1.2, 1.2c0, and 1.3a0 and neither syncdb nor south migrations appear to be working and running the South migration command fails with:

``` You have no migrations for the 'registration' app. You might want some. ```

'register' is added to INSTALLED_APP.

Here's the output from syncdb:

```
$ python manage.py syncdb
Syncing...
Creating tables ...
Installing custom SQL ...
Installing indexes ...
Installed 10 object(s) from 1 fixture(s)

Synced:
 > grappelli
 > radmin
 > admin_timeline
 > singleton_models
 > adminactions
 > django.contrib.admindocs
 > django.contrib.admin
 > south
 > django.contrib.auth
 > django.contrib.contenttypes
 > django.contrib.sessions
 > django.contrib.sites
 > django.contrib.messages
 > django.contrib.markup
 > django.contrib.staticfiles
 > captcha
 > honeypot
 > publications
 > django.contrib.comments
 > debug_toolbar
 > djcelery_email
 > rest_framework

Not synced (use migrations):
 - reversion
 - explorer
 - sbgrid_contacts
 - sbgrid
 - custom
 - flatpages
 - django_extensions
 - registration
(use ./manage.py migrate to migrate these)

```

And the output from running the South migrate command:

```
$ python manage.py migrate

Running migrations for reversion:
- Nothing to migrate.
 - Loading initial data for reversion.
Installed 0 object(s) from 0 fixture(s)
Running migrations for explorer:
- Nothing to migrate.
 - Loading initial data for explorer.
Installed 0 object(s) from 0 fixture(s)
Running migrations for custom:
- Nothing to migrate.
 - Loading initial data for custom.
Installed 0 object(s) from 0 fixture(s)
Running migrations for flatpages:
- Nothing to migrate.
 - Loading initial data for flatpages.
Installed 0 object(s) from 0 fixture(s)
Running migrations for django_extensions:
- Nothing to migrate.
 - Loading initial data for django_extensions.
Installed 0 object(s) from 0 fixture(s)
? You have no migrations for the 'registration' app. You might want some.
```

This is in a virtualenv and I can see the migration files in the virtualenv:

```
 ls ~/.virtualenvs/sbgrid_v2/lib/python2.7/site-packages/registration/migrations/
[...]

PR #94, opened 2015-08-27
Fix south migrations, previous checkin had models that don't exist in…

mick-t:
See

* https://github.com/macropin/django-registration/issues/87#issuecomment-134452825

Mauler's commit has references to models that don't exist, and I presume are custom models for his own project. This should resolve that.

### DIFF
--- a/registration/south_migrations/0001_initial.py
+++ b/registration/south_migrations/0001_initial.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-from south.utils import datetime_utils as datetime
+import datetime
 from south.db import db
 from south.v2 import SchemaMigration
 from django.db import models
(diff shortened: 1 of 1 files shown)


dicato:
@mick-t I apologize this has sat. Would you be able to rebase this branch so it applies cleanly and retest the migration to ensure it works?

dicato:
After looking through pull request history, I'm going to close this as support to Django 1.7 and South was dropped in #117.

@mick-t if this causes you problems, let's discuss. :smile:

mick-t:
Sorry for not replying soon -- I was away for a few weeks with limited Internet access and am finally back online. :)

I'm still using Django 1.5 with plans to upgrade to 1.9. In the meantime I still need to use Django-registration -- how would you suggest I proceed? 

I can rebase my branch if you like?

dicato:
@mick-t no worries!

I'm honestly not sure the best approach. It was a conscious decision to drop support for Django older than 1.7. If I was running a Django 1.5 project, I'd likely fork this project and add south migrations back.

I'd like for others to weigh in on this as well.

mick-t:
Your fork suggestion is what I was thinking of doing -- I guess that's what
I'll do. Thanks.

PR #97, opened 2015-09-04
Removed direct inheritance from `base.html`

gbdlin:
Removed direct inheritance from `base.html` on activation templates,
instead of direct inheritance, activation templates now extends
`registration/activation_base.html`, and that template extends
`base.html`. That way we can change what template is extended by
activation templates, like it was made for other templates already.

### DIFF
--- a/registration/templates/registration/activate.html
+++ b/registration/templates/registration/activate.html
@@ -1,4 +1,4 @@
-{% extends "base.html" %}
+{% extends "registration/activation_base.html" %}
 {% load i18n %}
 
 {% block content %}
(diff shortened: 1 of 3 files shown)


areski:
@gbdlin just checking if https://github.com/macropin/django-registration/pull/108 is working for you and if so, let's close this ticket. Thanks

dicato:
@areski thanks for following up!

gbdlin:
Of course, #108 will work for me. I just assumed that those templates aren't extending registration_base for some reason. If there is no reason for that, #108 will be better

dicato:
Thanks @gbdlin for confirming. Closing this. Thank you both!

PR #108, opened 2015-09-21
fix extends base

areski:


areski:
Tests brake on python 3.2, maybe we should rerun them

dicato:
This appears to be a duplicate change of #97.

gbdlin:
@dicato not entirely, my edit is creating separate base file for activation templates. I thought registration_base.html wasn't used here for some reason. So it is this pull request or #97

areski:
In my usercase, it's practical for me to overload registration_base.html so that I don't have to change the other templates.
I'm happy with @gbdlin solution too, that will work too

areski:
@dicato Do you have some recommendation to push this forward?

dicato:
Honestly, I don't have a strong preference for either solution. I do think simplicity is better, which makes me lean towards this pull request. I'm not sure if it solves what @gbdlin set out to solve with #97. If it does, then let's push with this pull request.

@areski can you look into why the tests are failing and address accordingly?

areski:
Ok that's error it's a tricky one, the test passing aren't related to my PR but tests will stop to pass on Python 3.2 and you can see you have the same problem on all the new PR
The culprit is https://github.com/nedbat/coveragepy/blob/master/coverage/annotate.py#L81
    dest.write(u'  ')

u is not supported in Python 3.2 but it's for all the recent version.

A lot of packages are dropping support for Python 3.2, I would advice to do so, eg https://github.com/pyca/cryptography/issues/1809

areski:
No Py32 here https://github.com/nedbat/coveragepy/blob/master/.travis.yml
So let's drop 3.2 support :)

dicato:
@areski thanks! I agree with dropping Python 3.2 support.

PR #111, opened 2015-09-24
improve README.rst

areski:


areski:
just small improvements on the readme, cc/ @dicato

dicato:
I like it! I have two thoughts for feedback:

* should we delete the `INSTALL` file?
* should we shuffle the order of installation instructions to promote installing from PyPI with pip as the preferred option?

areski:
Oups forgot to delete the INSTALL, that was my intention.
I can't agree more on your second point!

Thanks

dicato:
Thanks @areski :+1: Merging!

areski:
Thanks :)

PR #112, opened 2015-09-25
Coverage configuration

joshchandler:
I've added a [`.coveragerc` configuration file](http://coverage.readthedocs.org/en/latest/config.html) that omits the `migrations` and `south_migrations` folders.  They aren't explicitly covered in tests, and shouldn't be included in code coverage. 

I set the branch coverage measurement to True, but coveralls.io doesn't seem to have that feature, yet.  If you wanted to see branch coverage, it'd have to be done locally.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,5 @@
 [run]
+branch = True
 omit = 
   */migrations/*
   */south_migrations/*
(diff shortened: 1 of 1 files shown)


dicato:
Thanks! There actually are ways to test migrations in Django. For example, see [django_migration_test_case](https://github.com/plumdog/django_migration_testcase).

joshchandler:
Nice, that could prove useful one day.  Thanks

PR #113, opened 2015-09-25
Some new form combinations

joshchandler:
I've created some new forms by combining features from the already existing forms.  Both the new forms just add a terms of service agreement element to the 2 forms with customized email elements.  

I plan on creating a `RegistrationFormUniqueAndNoFreeEmail` class, but I didn't want to create that one yet until I got feedback on this pull request.


dicato:
Do you prefer for these to be in django-registration instead of having to subclass and create your own in the Django project you are using them in? I personally do not have a strong preference and have answered at least one question on how to do this type of multi-inheritance in order to achieve functionality like this.

joshchandler:
I have no preference either way.  I have used the unique email and tos combination in multiple projects and thought others would find it useful.  It could, at least, be worth pointing out in the forms section of the docs the possibilities and limitations of multi-inheritance of these forms.

dicato:
I think it's worth including. Would you be willing to update the documentation to describe this? You can add them to the [forms docs](https://github.com/macropin/django-registration/blob/master/docs/forms.rst)

joshchandler:
I think it would suffice to only have this section in the docs instead of creating all these different form variations, as they don't necessarily add anything substantial to the project.

areski:
Good with me, documentation is clear and work fine on the latest version!
+1

dicato:
Thank you @joshchandler and @areski!

PR #115, opened 2015-10-06
resend activation email logic + extensibility

falcon1kr:
1. resend activation email logic added (changes the activation code to a new one)
2. better extensibility of the RegistrationProfile by adding profile_info parameter to create_inactive_user and create_profile manager methods
3. making transactions atomic where both user and profile info are modified together

resend logic needs to be surfaced to default views (I didn't need to add this since I was using my own views for my project)


dicato:
I believe this just need to be rebased on master to ensure the tests pass.

areski:
I added an extra patch on your PR, see https://github.com/macropin/django-registration/pull/137
The patch is tiny, it's just to use your new method in the admin.

Thanks for the hard work!

dicato:
Merged #137 instead; thank you both :smile:

areski:
All credits to @falcon1kr  :) nice addition!

PR #117, opened 2015-10-28
drop support for Python 2.6 and Django < 1.7, as discussed in #92

jamesturk:
(also adds support for 1.9 and fixes a DeprecationWarning for 1.10)


macropin:
Thanks for this. Looks comprehensive.

PR #133, opened 2015-12-12
Update faq.rst

vocaviking:
Edit concerning a problem I had, and spend around 2 hours searching for the solution.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -130,6 +130,11 @@ Troubleshooting
     general, it is always a good idea to follow normal Python
     practices for installing and using Django applications).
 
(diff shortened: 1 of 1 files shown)


dicato:
@vocaviking could you explain the issue you had? Which templates where causing you issues?

vocaviking:
I wanted to write custom templates for all the views. To do so, I put my template files in the usual places. However for some views, django used the admin templates instead of mine for no apparent reason. As it turns out, you can fix it, by editing the INSTALLED_APPS in settings.py
(Hope this makes sense to you)

dicato:
Was this happening specifically for password reset templates? I can see that happening.

vocaviking:
Indeed.

PR #134, opened 2015-12-14
Use Django 1.9 instead of 1.9b1 in travis

awbacker:
Figured I would start with something small, but necessary.  I am trying to upgrade to Django 1.9, and since the beta is officially over its probably a good idea to feed travis the latest django.  

Builds successfully.  https://travis-ci.org/awbacker/django-registration


dicato:
:boom: thanks! That's helpful :smile:

awbacker:
Haha, and tiny!

PR #136, opened 2015-12-17
Fix deprecated class NoArgsCommand class.

jcamiel:
Solve the warning RemovedInDjango110Warning: NoArgsCommand class is deprecated and will be removed in Django 1.10. Use BaseCommand instead, which takes no arguments by default.


dicato:
Thank you! Works for me. :+1:

PR #141, opened 2016-01-12
Add alternatives to README.

dicato:
I believe it's worth adding alternatives to django-registration in the README. django-registration provides a succinct user registration package, but user registration is only one component of self-servicing user management. As you begin to support other forms of authentication, email as username, two-factor, etc, you are required to customize django-registration extensively. Because of this, I have migrated multiple projects to [django-allauth](http://www.intenct.nl/projects/django-allauth/) with success.

Does anyone have any other alternative packages that should be mentioned?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -60,9 +60,13 @@ functional installation of Django 1.7 or newer.
 Alternatives
 ------------
 
-`django-allauth <http://www.intenct.nl/projects/django-allauth/>`_ is an
-alternative to django-regisitration-redux that provides user registration in
-addition to social authentication and email address management.
+`djangopackages.com <https://www.djangopackages.com/grids/g/registration/>`_
+has a comprehensive comparison of Django packages used for user registration
+and authentication.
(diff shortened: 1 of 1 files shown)


dicato:
@macropin & @areski any thoughts on this? I don't want to intentionally push people away from django-registration but:

* awareness to alternatives is always good
* I've personally stopped using django-registration in favor of django-allauth

areski:
I'm in favor of awareness so that peoples can decide for themselves which one they want to use.

I believe that there is cases where one need *simple* registration and doesn't want to use an extensive plugins but rather something small that do exactly what you want.

dicato:
@areski I concur! I tried to not prescribe any preference in the actual README. :+1:

mcrowson:
How about just linking to https://www.djangopackages.com/grids/g/registration/

dicato:
@mcrowson thanks for that! That's a great idea.

PR #143, opened 2016-01-13
Template Settings Change for Django 1.10

mcrowson:


dicato:
@mcrowson do you have a link to the related changes in Django? It would help with reviewing. Thanks!

mcrowson:
https://docs.djangoproject.com/en/1.9/ref/templates/upgrading/

mcrowson:
Here is the deprecation timeline showing that the old TEMPLATE_* settings will be removed. https://docs.djangoproject.com/en/1.9/internals/deprecation/#deprecation-removed-in-1-10

mcrowson:
The URL change is for the following 1.10 deprecation:
"Support for string view arguments to url() will be removed."

dicato:
Thanks. It looks good to me! I also tested locally with Django 1.9 and the test app and everything works as expected.

In order to improve this project's release cycle, could you update the `CHANGELOG` for this compatibility fix?

mcrowson:
Sure, do you want it under 1.3? I don't see any conventions for a 1.3.1 or anything. Or are you saying to start a 1.4 section and that this will be included in the 1.4 release?

dicato:
I'd say start a 1.4 section with a "TBD" date. I think that's the simplest approach.

dicato:
Thank you!

PR #145, opened 2016-01-22
Used migrate instead of syncdb

mohan3d:
syncdb is deprecated since version 1.7
https://docs.djangoproject.com/en/1.7/ref/django-admin/#syncdb


dicato:
Thanks!

PR #154, opened 2016-02-19
Add makefile to help developers get started

joshblum:
If you guys want to merge this, we can also add some docs to the `README` to let people know about it.

What do you guys typically do to manage the different versions of python/Django for testing?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,5 @@
+flake8
+coveralls
+Sphinx
+Django>1.7,<1.9
+invoke==0.12.2
--- a/tasks.py
+++ b/tasks.py
@@ -3,16 +3,12 @@ from invoke import task
 
 
 @task
-def clean(docs=False, bytecode=True, extra=''):
-    patterns = ['build']
-    if docs:
-        patterns.append('docs/_build')
-    if bytecode:
-        patterns.append('**/*.pyc')
-    if extra:
-        patterns.append(extra)
-    for pattern in patterns:
-        run("rm -rf %s" % pattern)
+def clean(all=False):
+    if all:
+        flag = "--all"
+    else:
+        flag = ""
+    run("python setup.py clean {}".format(flag))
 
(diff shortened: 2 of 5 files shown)


dicato:
Personally, I have a virtualenv for this project and simply test locally against the latest Python 2.7 and Django 1.x. I let Travis matrix the tests against other versions for me. I tend to run `pip install -e .` and `python setup.py test` to test locally.

I'm not strongly opposed to a `makefile`, but being a Python project I'd advocate for using [invoke](http://www.pyinvoke.org/) as well as adding a `requirements.txt` file for familiarity.

joshblum:
+1 to the `requirements.txt`, i didn't know if you guys didn't want a file since it's just packages for dev. I've never used invoke, but I'll take a look after work and push an update

dicato:
Sounds great! Thanks for taking the time to make working on this project easier. :smile:

joshblum:
@dicato take a look and let me know what you think!

tasks.py
```diff
@@ -0,0 +1,32 @@
+from invoke import run
+from invoke import task
+
+
+@task
+def clean(docs=False, bytecode=True, extra=''):
```
dicato: Does it make sense to call `python setup.py clean` (or optionally `clean --all`) instead of building commands manually? It's likely less prone to user error.

joshblum: Much better, didn't know this was already implemented!

dicato:
Looks good to me!

requirements.txt
```diff
@@ -0,0 +1,5 @@
+flake8
+coveralls
+Sphinx
+Django>1.7,<1.9
```
dicato: Whoops, just noticed this `<1.9` which will stop 1.9.x from being installed! I'll fix it. :smile:

joshblum: Oops! Good catch thanks :)

Issue #158, opened 2016-03-06
Include url to resend activation email


joshblum:
Is there a reason that an endpoint to resend the activation email doesn't exist already? It would be nice to have this so a user can hit an endpoint if they lost the email.

Essentially a wrapper around this function:
https://github.com/macropin/django-registration/blob/master/registration/admin.py#L44

If you guys think it would be good I don't mind implementing :)

dicato:
Would it be an authenticated or anonymous view (or both)?

I verified for my own knowledge that `resend_activation_key` creates a new key only if a corresponding profile with said email exists and has not already been activated or expired; so that's good - https://github.com/macropin/django-registration/blob/f25a0b63da92a49ed07d5be39eafa636a5562295/registration/models.py#L137

joshblum:
I think anon view since authenticated users shouldn't need to register again right?

joshblum:
@lydiagu here is the issue for reference

dicato:
@joshblum oh..yea you are right. I totally confused registered with "verified email" in my mind :see_no_evil:

joshblum:
closed by #161

PR #161, opened 2016-03-23
Implement endpoint to resend activation email.

lydiagu:
Here's a stab at implementing #158


dicato:
A new feature *and* tests! Thank you very much :smile:

registration/backends/default/views.py
```diff
@@ -139,3 +140,18 @@ def activate(self, *args, **kwargs):
...
+
+
+class ResendActivationView(BaseResendActivationView):
+    def resend_activation(self, *args, **kwargs):
+        """
+        Given an email, look up user by email and resend activation key if user
+        is not already activated or previous activation key has not expired.
+        """
+        site = get_current_site(self.request)
+        email = kwargs.get('email', '')
```
dicato: What happens when there is no `email` in kwargs and empty string is used?

lydiagu: `resend_activation_mail` looks for a RegistrationProfile with `email`. An empty string would probably not match any RegistrationProfiles, and the function returns False. This will show the resend_activation.html template. Is there a standard for returning bad request instead?

dicato: I think that makes sense, albeit following the code is not super clear. I'm not sure of a simpler implementation though.

dicato:
@lydiagu sorry for the delays! I've finally been able to try this feature with the test app and give it a proper review. Here'a a couple things I've noticed:

1) the different error messages reveal if an email address has an associated account. Trying to resend activation with an empty `email` renders this page and redirects back to the resend view:

screen shot 2016-04-09 at 9 28 29 am

...trying to resend to a non-existent email renders this page and redirects back to the resend view:

screen shot 2016-04-09 at 9 28 51 am

...and successfully resending activation renders this page and redirects to complete:

screen shot 2016-04-09 at 9 28 51 am

I think it makes sense to render a consistent template with a message similar to:

> We have sent an email to {{ email }} with further instructions.

...in both scenarios where we have found an associated account/registration and not found an associated account/registration. This helps prevent being able to enumerate accounts within the application using django-registration.

2) resend complete page uses `email` in the template which is not available in the context

3) I suspect, though am not positive, that the implementation and user experience would be improved by using a `FormView`. Users are expected to type in their email address, so a form makes sense from that perspective. Additionally, a `FormView` will make it easier to pass the email that was entered between views for rendering in templates.

Does this make sense?

lydiagu:
Thanks @dicato for the feedback! I'll change to:

- using a FormView for getting the email address
- if the form is valid, redirect to an Activation Resent Success page regardless if the email is valid or not

One question about the codebase: what part of the implementation should go in registration/views vs in registration/backends/default/views? I roughly tried to follow the existing pattern, but I'm not sure I understand what should go where.

Thanks!

dicato:
@lydiagu I think the pattern you followed for separating the code is good. I'd keep it that way as it puts the logic in the same place as the rest of that activation workflow.

It's certainly a little confusing. It makes me wonder how many people are actually extended the abstracted class.

lydiagu:
Is the general idea that the base class calls the function that actually does the work, for example, self.resend_activation, and handles redirect, but self.resend_activation and self.get_success_url would be implemented in the subclass?

dicato:
In general, I think the business logic, such as resending the activiation, should be in the subclass (`backend.default.views`) and the stuff like success urls, redirects, all the little "Django things" can go in the base class. 

I *think* that's what you said :smile:

lydiagu:
Hi! Sorry for the delay. I changed the code to use a FormView for requesting a resend of the activation key. Also fixed the rendering of the email address in the response.

dicato:
Looks like the build is failing due to some simple flake issues!

I hope to review this code this week!

registration/models.py
```diff
@@ -142,6 +142,7 @@ def resend_activation_mail(self, email, site, request=None):
             profile = self.get(user__email=email)
         except ObjectDoesNotExist:
             return False
+        # TODO: Catch multiple objects returned exception?
```
dicato: @lydiagu does this TODO still need to be addressed?

dicato:
Other than the aforementioned TODO, this is ready to be merged! Thanks @lydiagu

registration/models.py
lydiagu: Yes, I added this todo because I don't think email is a necessarily a unique field on user. Do you think it will always be unique? If there are multiple users with the same email address, should they all be resent an activation? If we want to change this to a filter instead of a get, I can do that in a separate PR.

dicato: That makes sense! Thanks for the clarification.

You are right. Emails are not unique by default in Django and django-registration. This is beyond ready to merge :smile: so let's get this in into master and I will file an issue to discuss the email uniqueness TODO and we can discuss.

lydiagu: Great thanks!

Issue #85, opened 2015-08-08
django.core.urlresolvers.NoReverseMatch: Reverse for 'password_reset_confirm'


ishaq:
**Django:** 1.8
**Python:** 3.4

When a user tries to reset password, it throws exception:

```
django.core.urlresolvers.NoReverseMatch: Reverse for 'password_reset_confirm' with
    arguments '()' and keyword arguments '{'token': '444-9887c242917d1b6b7368',
    'uidb64': b'Mg'}' not found. 0 pattern(s) tried: []
```

[This thread](https://stackoverflow.com/questions/19985103/django-1-6-and-django-registration-built-in-authentication-views-not-picked-up) on Stackoverflow solved it for me.

>The reason for this error is that the _django.contrib.auth.views_ use different url names than the _registration.auth_urls_ ones. To patch this problem, override the default urls until _django-registration-redux_ gets updated for django 1.6, and use the same names as Django.

```python
from django.contrib.auth import views as auth_views

urlpatterns = patterns('',

      #override the default urls
      url(r'^password/change/$',
                    auth_views.password_change,
                    name='password_change'),
      url(r'^password/change/done/$',
                    auth_views.password_change_done,
                    name='password_change_done'),
      url(r'^password/reset/$',
                    auth_views.password_reset,
                    name='password_reset'),
      url(r'^password/reset/done/$',
                    auth_views.password_reset_done,
                    name='password_reset_done'),
      url(r'^password/reset/complete/$',
                    auth_views.password_reset_complete,
                    name='password_reset_complete'),
      url(r'^password/reset/confirm/(?P<uidb64>[0-9A-Za-z]+)-(?P<token>.+)/$',
                    auth_views.password_reset_confirm,
                    name='password_reset_confirm'),

      #and now add the registration urls
      url(r'^accounts/$', include('registration.backends.default.urls')),
)
```

dicato:
@ishaq I usually include auth URLs in the same way you describe.

Did you see [this documentation about registration.auth_urls](https://django-registration-redux.readthedocs.org/en/latest/faq.html#configuration)? Does it help?

ishaq:
@dicato thanks for the link to the documentation. I had seen it though, It was not helpful because it doesn't say password reset function will not work unless one puts additional patterns in url conf. It was only after seeing the exception that I realised some url patterns are missing.

[...]

PR #166, opened 2016-03-28
updating docs fix#85

aj07mm:
adding comment for prob on:
https://github.com/macropin/django-registration/issues/85#issuecomment-133068403


dicato:
Thanks!

Issue #171, opened 2016-04-20
Change account activation logic


joshblum:
Right now the expiration date on a `RegistrationProfile` is determined by `user.date_joined`, however in our workflow we sometimes create users before they actually go through the registration flow. As an alternative would it be possible to add a field to `RegistrationProfile` so that the activation key time can be independent of the `user.date_joined`. Happy to submit a PR, just wanted to run the strawman by you guys first :)

dicato:
I don't see any obvious issues with the approach. You'll obviously need a schema migration. You'll probably need a data migration as well.

joshblum:
Hmm a simpler alternative could be to update the `user.date_joined` in `RegistrationProfile.create_inactive_user`, this is our temporary workaround. Thoughts?

dicato:
merged!

PR #172, opened 2016-04-22
Reset user.date_joined to ensure a full activation interval window

joshblum:


joshblum:
Addresses #171

joshblum:
@dicato thoughts on this approach?

dicato:
Interesting! I think it's a simple and clever approach. Do you see any side effects for people who use the User's join date in other logic?

joshblum:
Hmm it is possible that others would use this field in logic before registration, perhaps we can leave a note in the documentation and then people can subclass to change `create_inactive_user` if necessary?

dicato:
Alternatively, we could decouple the date joined with the registration expiry date. We would add a field to `RegistrationProfile` for this. Thoughts?

joshblum:
The `user.date_joined` is definitely simpler (which is why I decided to give it a try vs. the data migration) if you think changing the date is a bad idea we can decouple but I'm not sure if it will mess with anything for anyone

dicato:
I honestly don't feel strongly either way. I think under normal circumstances this implementation won't cause any unexpected issues. Maybe someone else has input?

joshblum:
Let's land before the next release if no one comments?

dicato:
👍 absolutely!

PR #175, opened 2016-05-05
Update 0003_migrate_activatedstatus.py

yeago:
save some memory, yah


dicato:
Ha, did this migration causes you issues which led to you making this change?

Issue #109, opened 2015-09-21
Manual validation from the admin


areski:
This is proposition for a new feature, the idea is that the admin have to manually accept the new registered user before they are allowed to register.

So after registration and mail validation the user, get on a view with "Your account is being review, we will confirm it shortly!"

Maybe a similar feature can be achieve with a workaround?

dicato:
Considering I've built that exact workflow in the past, I agree it's a useful feature.

Could you take a look at [django-allauth](https://github.com/pennersr/django-allauth) and determine whether that workflow is already supported?

areski:
As far as I could see that workflow doesn't seem to be supported in django-allauth

dicato:
Thanks for doing the research.

If you work on a pull request, I'm happy to provide input and review. I will say that after using that workflow in production, we eventually moved away from it as it created a very high barrier of entry for adoption of our services, which was leading to us losing users. Just food for thought. There are absolutely use cases for this.

areski:
I understand that  in most cases this is not the willing behavior. Usually I don't want this, but in one of my scenario -I work in VoIP and payment- we want to run a manual verification on new user to reduce fraud risks.

I will send a PR as soon as this comes back to my priority list.

lucbran:
This is something I am also interested...any update?

joshblum:
@areski are you still interested in working on this?

areski:
@joshblum I didn't start anything in that area, but I would love to see a PR so if you want to take over and propose a patch please go ahead.

joshblum:
@areski I don't have time right now to submit a patch myself, but just wanted to see if the issue was current. Since the feature is pretty specific, I think ti would be best to close the issue for now and reopen if someone submits a PR.

dicato:
@joshblum I'd leave the issue open and slap the "help wanted" label on it since the feature request is still valid.

joshblum:
sounds good!

joshblum:
closed by #184

PR #184, opened 2016-06-06
Add 3-step registration workflow by implementing another registration backend

sergafts:
This includes the changes required to implement a 3-step registration workflow.

1) User registers an account - is emailed with an activation token
2) User activates the account - site administrators are emailed in order to approve the user's newly created account
3) Administrators approve user's account - User is emailed and can now log in.

It contains a new registration backend which has been tested with all the previous tests and some newly added to test the new functionality.

For more information refer to the commit message. If the changes are not clear enough, I would be happy to provide documentation too. I have not done so yet as I would like some feedback first.

I am available to answer any possible questions. Thank you.


joshblum:
Addresses #109

joshblum:
@sergafts Thanks for implementing!! High level I think you should clean up the code to not repeat the logic already defined in the default backend. Just subclass the respective objects where necessary to add the approval logic. I can give some more detailed feedback after this, it's hard to read such a large diff with the repeated code. But again, thanks for submitting a full feature with tests!

sergafts:
@joshblum Thank you for your comments, I am processing them.

This was my initial thought too. This is why ``SupervisedRegistrationProfile`` subclasses ``RegistrationProfile``. As far as the managers are concerned, by subclassing ``RegistrationManager`` I would only be able to reuse methods ``create_inactive_user()``, ``create_profile()`` and ``resend_activation_email()``, as all other methods have changed. Since you think it is worth it, I am happy to implement it. It seemed cleaner (could be just me though) to create a different class and perhaps repeat some code. I will work on it and get back to you. I also have some additional tests in mind to help maintain coverage levels high.

sergafts:
@joshblum I would like some feedback if possible to help me continue with the process.

registration/backends/enhanced/urls.py
```diff
@@ -0,0 +1,73 @@
+"""
```
joshblum: @dicato do you have an opinion between sharing URLs between backends? All of these URLs are the same from the `default` backend, except the `approve` related ones. Should we just append to the `default` urls or leave this file as is?

joshblum:
can you rename the backend to something more descriptive such as `manual_approval` or `admin_approval` or something similar instead of `enhanced`?

registration/backends/enhanced/views.py
```diff
@@ -0,0 +1,172 @@
...
+            signals.user_activated.send(sender=self.__class__,
+                                        user=activated_user,
+                                        request=self.request)
+        return activated_user
+
+    def get_success_url(self, user):
+        return ('registration_activation_complete', (), {})
+
+
+class ApprovalView(BaseApprovalView):
```
joshblum: pending @dicato comments about the urls file, i believe that all views except this one should be inherited from the `default` backend. There is no need to repeat this code and then have to propagate bug fixes between the two.

registration/models.py
```diff
@@ -376,3 +377,364 @@ def send_activation_email(self, site, request=None):
...
+                    return False
+
+            if not profile.activation_key_expired():
+                if not profile.user.is_active and not profile.activated:
+                    self.send_admin_approve_email(profile.user, site)
+
+                # do not set ``User.is_active`` as True. This will be set
+                # when a site administrator approves this account.
+                profile.activated = True
+                with transaction.atomic():
```
joshblum: i don't believe that this is necessary since we are only saving a single object

registration/models.py
```diff
@@ -376,3 +377,364 @@ def send_activation_email(self, site, request=None):
...
+                # return False as the User has not yet been approved by
+                # a site administrator or has been deactivated.
+                if profile.user.is_active:
+                    return True
+                else:
+                    return False
+
+            if not profile.activation_key_expired():
+                if not profile.user.is_active and not profile.activated:
+                    self.send_admin_approve_email(profile.user, site)
```
joshblum: can you modify the `RegistrationManager` to perform this logic in a private method, such as `_activate_user`. This way you can just subclass this one method and perform the admin approval logic and reduce the repeated code

registration/models.py
```diff
@@ -376,3 +377,364 @@ def send_activation_email(self, site, request=None):
...
+                if get_profile:
+                    return profile
+                else:
+                    return True
+        return False
+
+    def admin_approve_user(self, profile_id, site, get_profile=False, request=None):
+        """
+        Approve the ``SupervisedRegistrationProfile``
+        object with the given ``id``.
```
joshblum: id->profile_id

registration/models.py
```diff
@@ -376,3 +377,364 @@ def send_activation_email(self, site, request=None):
...
+        after approving.
+
+        If the id is not valid, return ``False``.
+
+        If the id is valid but the ``User`` is already active,
+        return ``User``.
+
+        If the id is valid but the ``SupervisedRegistrationProfile``
+        object is not activated, return ``False``.
+
```
joshblum: remove whitespace

registration/models.py
```diff
@@ -376,3 +377,364 @@ def send_activation_email(self, site, request=None):
...
+
+        """
+        try:
+            profile = SupervisedRegistrationProfile.objects.get(id=profile_id)
+            if profile.activated:
+                if profile.user.is_active:
+                    return profile.user
+
+            # If the user has not activated his profile the admin should
```
joshblum: their profile

registration/models.py
```diff
@@ -376,3 +377,364 @@ def send_activation_email(self, site, request=None):
...
+            # not be able to approve his account (at least not following
+            # this process)
+            if profile.activated:
+                profile.user.is_active = True
+            else:
+                return False
+
+            with transaction.atomic():
```
joshblum: not needed

registration/models.py
```diff
@@ -376,3 +377,364 @@ def send_activation_email(self, site, request=None):
...
+
+            if get_profile:
+                return profile
+            else:
+                return profile.user
+        except self.model.DoesNotExist:
+            return False
+
+    def send_admin_approve_email(self, user, site, request=None):
+
```
joshblum: remove whitespace

registration/models.py
```diff
@@ -376,3 +377,364 @@ def send_activation_email(self, site, request=None):
...
+        These templates will each receive the following context
+        variables:
+
+        ``user``
+            The new user account
+
+        ``user_id``
+            The id of the associated``SupervisedRegistrationProfile``
+            object.
```
joshblum: is it a `user_id` or `profile_id`?

registration/models.py
```diff
@@ -376,3 +377,364 @@ def send_activation_email(self, site, request=None):
...
+        )
+        admin_approve_email_html = getattr(
+            settings,
+            'ADMIN_APPROVAL_EMAIL_HTML',
+            'registration/admin_approve_email.html'
+        )
+
+        ctx_dict = {}
+        if request is not None:
+            ctx_dict = RequestContext(request, ctx_dict)
```
joshblum: Follow the pattern here so we don't need to use `RequestContext` anymore https://github.com/macropin/django-registration/pull/194

registration/models.py
```diff
@@ -376,3 +377,364 @@ def send_activation_email(self, site, request=None):
...
+                             settings.DEFAULT_FROM_EMAIL)
+        message_txt = render_to_string(admin_approve_email_body,
+                                       ctx_dict)
+
+        # try:
+        admins = getattr(settings, 'ADMINS')
+        admins = [admin[1] for admin in admins]
+        if not admins:
+            raise ImproperlyConfigured(
+                'Using the enhanced registration backend'
```
joshblum: propagate the new name here as well :)

registration/models.py
```diff
@@ -376,3 +377,364 @@ def send_activation_email(self, site, request=None):
...
+                message_html = render_to_string(
+                    admin_approve_email_html, ctx_dict)
+            except TemplateDoesNotExist:
+                pass
+            else:
+                email_message.attach_alternative(message_html, 'text/html')
+
+        email_message.send()
+
+    def delete_expired_users(self):
```
joshblum: you can remove this method since you are a subclass of `RegistrationManager`

registration/models.py
```diff
@@ -376,3 +377,364 @@ def send_activation_email(self, site, request=None):
...
+
+@python_2_unicode_compatible
+class SupervisedRegistrationProfile(RegistrationProfile):
+
+    # Same model as ``RegistrationProfile``, just a different
+    # Manager to implement the extra functionality required
+    # in admin approval
+    objects = SupervisedRegistrationManager()
+
+    def __str__(self):
```
joshblum: not need, verbatim to superclass

registration/models.py
```diff
@@ -376,3 +377,364 @@ def send_activation_email(self, site, request=None):
...
+        admin_approve_complete_email_body = getattr(
+            settings, 'APPROVAL_COMPLETE_EMAIL_BODY',
+            'registration/admin_approve_complete_email.txt')
+        admin_approve_complete_email_html = getattr(
+            settings, 'APPROVAL_COMPLETE_EMAIL_HTML',
+            'registration/admin_approve_complete_email.html')
+
+        ctx_dict = {}
+        if request is not None:
+            ctx_dict = RequestContext(request, ctx_dict)
```
joshblum: Follow the pattern here so we don't need to use RequestContext anymore #194 :)

registration/models.py
```diff
@@ -376,3 +377,364 @@ def send_activation_email(self, site, request=None):
...
+        # can overwrite some of the values like user
+        # if django.contrib.auth.context_processors.auth is used
+        ctx_dict.update({
+            'user': self.user,
+            'site': site,
+        })
+        subject = (getattr(settings, 'REGISTRATION_EMAIL_SUBJECT_PREFIX', '') +
+                   render_to_string(
+                       admin_approve_complete_email_subject, ctx_dict))
+        # Email subject *must not* contain newlines
```
joshblum: can you abstract this as a send_email helper function? this code is repeated in the `SupervisedRegistrationManager` and `RegistrationProfile` classes

registration/templates/registration/activation_complete_admin_pending.html
```diff
@@ -0,0 +1,22 @@
+{% extends "registration/registration_base.html" %}
```
joshblum: can you provide screenshots of these?

registration/templates/registration/admin_approve_complete.html
```diff
@@ -0,0 +1,20 @@
+{% extends "registration/registration_base.html" %}
+{% load i18n %}
+
+{% block title %}{% trans "Account Approved" %}{% endblock %}
+
+{% block content %}
+<p>
+    {% trans "The user's account is now approved." %}
+
```
joshblum: remove line

registration/templates/registration/admin_approve_complete.html
```diff
@@ -0,0 +1,20 @@
...
+</p>
+{% endblock %}
+
```
joshblum: remove line

registration/templates/registration/activation_complete_admin_pending.html
```diff
@@ -0,0 +1,22 @@
...
+
+{% block content %}
+<p>
+    {% trans "Your account is now activated." %}
+    {% if not user.is_authenticated %}
+    {% trans "Once a site administrator activates your account you can login." %}
+    {% endif %}
+</p>
+{% endblock %}
+
```
joshblum: remove line

registration/templates/registration/admin_approve.html
```diff
@@ -0,0 +1,18 @@
+{% extends "registration/registration_base.html" %}
+{% load i18n %}
+
+{% block title %}{% trans "Approval Failure" %}{% endblock %}
+
+{% block content %}
+<p>{% trans "Account Approval failed." %}</p>
+{% endblock %}
+
```
joshblum: remove line

registration/templates/registration/admin_approve_complete_email.html
```diff
@@ -0,0 +1,26 @@
...
+<html lang="en">
+
+<head>
+    <title>{{ site.name }} {% trans "admin approval" %}</title>
+</head>
+
+<body>
+<p>
+    {% blocktrans %}
+    Your account is now approved. You can log in.
```
joshblum: provide a link to the login page?

registration/templates/registration/admin_approve_complete_email.txt
```diff
@@ -0,0 +1,12 @@
+{% load i18n %}
+{% blocktrans with site_name=site.name %}
+Your account is now approved. You can log in.
```
joshblum: login link?

registration/templates/registration/admin_approve_email.html
```diff
@@ -0,0 +1,35 @@
...
+<html lang="en">
+
+<head>
+    <title>{{ site.name }} {% trans "registration" %}</title>
+</head>
+
+<body>
+<p>
+    {% blocktrans with site_name=site.name %}
+    The following user (NEED TO ADD USER HERE) has asked to register an account at
```
joshblum: `{{ user.username }}`?

registration/templates/registration/admin_approve_email.html
```diff
@@ -0,0 +1,35 @@
...
+</p>
+<p>
+    {% blocktrans %}
+    To approve this, please click the following link.
+    {% endblocktrans %}
+</p>
+
+<p>
+    <a href="http://{{site.domain}}{% url 'registration_admin_approve' user_id %}">
+        {{site.domain}}{% url 'registration_admin_approve' user_id%}
```
joshblum: `user_id %}` (add space)

registration/templates/registration/admin_approve_email.html
```diff
@@ -0,0 +1,35 @@
...
+<body>
+<p>
+    {% blocktrans with site_name=site.name %}
+    The following user (NEED TO ADD USER HERE) has asked to register an account at
+    {{ site_name }}.
+    {% endblocktrans %}
+</p>
+<p>
+    {% blocktrans %}
+    To approve this, please click the following link.
```
joshblum: why not just add the `a href` in this paragraph? So it would read something like `To approve this, please <a href=[...]>click here</a>`

registration/templates/registration/admin_approve_email.txt
```diff
@@ -0,0 +1,10 @@
+{% load i18n %}
+{% blocktrans with site_name=site.name %}
+The following user (NEED TO ADD USER HERE) has asked to register an account at
```
joshblum: `{{ user.username }}`

registration/tests/enhanced_backend.py
```diff
@@ -0,0 +1,258 @@
+import datetime
```
joshblum: if this code is all copied, please just subclass and set the `urls` attribute

registration/tests/models.py
```diff
@@ -424,3 +424,594 @@ def test_resend_activation_email_expired_user(self):
             site=Site.objects.get_current(),
         ))
         self.assertEqual(len(mail.outbox), 0)
+
+
+class SupervisedRegistrationModelTests(TestCase):
```
joshblum: same here, just add the tests for your new functionality and let the super class test the other methods

test_app/settings.py
```diff
@@ -23,7 +23,7 @@
```
joshblum: leave this unchanged

joshblum:
@sergafts left you some comments, largely about refactoring to avoid code duplication. Please let me know if you have any questions!! Sorry about the delay in reviewing, if I'm not responsive just @joshblum me and I'll take a look as soon as I have time :)

registration/models.py
sergafts: Sure, thanks.

registration/models.py
sergafts: I could, but I did not want to alter already existing code since I am implementing a new backend. But I guess it seems to make more sense this way.

registration/models.py
sergafts: Corresponding method in `RegistrationManager` has a difference. It executes `user.delete()` first and then `profile.delete()`. This failed during the tests, I was getting a `self.model.MultipleObjectsReturned( DoesNotExist: RegistrationProfile matching query does not exist)` error. I am guessing that since `RegistrationProfile` has a `ForeignKey` to the user when the user was deleted so was `RegistrationProfile` and deleting an object that extends it yielded this error. I am not 100% sure though, if you have any other ideas why this would happen I am open to hearing them.

registration/models.py
sergafts: Right

registration/models.py
sergafts: Sure, that would be nice. Implementing.

registration/templates/registration/admin_approve_complete_email.html
sergafts: Right

registration/templates/registration/admin_approve_complete_email.txt
sergafts: Right

registration/templates/registration/admin_approve_email.html
sergafts: Ah, that slipped through the cracks

registration/templates/registration/admin_approve_email.html
sergafts: Could be, seems simpler.

registration/models.py
sergafts: A `profile_id` actually.

sergafts:
> can you rename the backend to something more descriptive such as manual_approval or admin_approval or something similar instead of enhanced?

Sure. I wasn't set on the name anyway. I think admin_approval seems more in touch with the other naming conventions I took. 

Thank you for your (thorough) comments. Implementing!

registration/models.py
sergafts: @joshblum 
Actually I am not sure which logic you are talking about. Could you be a bit more specific?

test_app/settings.py
sergafts: If I leave this file unchanged it means there will be no value in `settings.ADMINS`. I have added a check in my code to raise `ImproperlyConfiguredError` if `settings.ADMINS` is not set. I guess we will not be needing that either.

registration/tests/enhanced_backend.py
sergafts: Actually it is not all copied as tests for the default backend use `RegistrationProfile` instead of `SupervisedRegistrationProfile`. I could alter the original `DefaultBackendViewTests` class to hold the registration profile it will use as a class variable and override it when subclassing, but I am not sure you would want that. If you think it would be better let me know.

joshblum: I think that setting a class variable would be best since most of the testing logic is identical (sorry I missed the import where you imported `SupervisedRegistrationProfile` as `RegistrationProfile`)

registration/models.py
joshblum: shouldn't the `RegistrationProfile` already be `@python_2_uinicode_compatible`?

test_app/settings.py
joshblum: `ImproperlyConfiguredError` would only be raised if your backend was being used correct? It's not needed by default

registration/models.py
joshblum: from line https://github.com/macropin/django-registration/blob/master/registration/models.py#L76 to https://github.com/macropin/django-registration/blob/master/registration/models.py#L82 in the `RegistrationManager` and then https://github.com/macropin/django-registration/pull/184/files#diff-950c315829fc87800c12895299e0c2f8R428 to https://github.com/macropin/django-registration/pull/184/files#diff-950c315829fc87800c12895299e0c2f8R435 in the `SupervisedRegistrationManager` make sense?

registration/models.py
joshblum: which test was failing? I'll take a closer look but if you had to switch two lines to subclass in the `RegistrationManager` it wouldn't be the end of the world

registration/models.py
sergafts: I guess not, since if I remove the declaration I have not explicitly defined `__str__() `. https://docs.djangoproject.com/en/dev/ref/utils/#django.utils.encoding.python_2_unicode_compatible
I guess inheriting this from the superclass is not enough.

test_app/settings.py
sergafts: Hmm. I see. My original thought was that when using the `admin_approval` backend it should be implied that you have set your `ADMINS`. How do you think I should ensure that?

registration/tests/enhanced_backend.py
sergafts: Seeing the code once more, I think perhaps I could follow the same logic with the class variable on probably most of the views in `registration/backends/admin_approval/views.py`. That is, altering the views on `registration/backends/default/views.py` to have a class variable and store the registration profile class and sublass them altering this value. I think this would be cleaner.

registration/templates/registration/activation_complete_admin_pending.html
sergafts: Would providing screenshots of the templates sent as emails be ok?

registration/models.py
sergafts: You can see more information here: https://titanpad.com/f7GUv9Hg5l

registration/models.py
sergafts: Actually I tried to implement this and I am not sure if it is possible. Since `SupervisedRegistrationManager.activate()` calls `send_admin_approve_email()`, it needs to know the current site. Right now it does so because it has an additional argument. But if its signature is changed to match `RegistrationManager.activate()` then this argument will not be available. 

Another way to get the current site would be from the `request` variable, but this is not available here as far as I am concerned.

The other method to do so would be to load the `Site` model and use `Site.objects.get_current()` but this is also not possible as doing so at import time causes this:
`AppRegistryNotReady("Models aren't loaded yet.")` ( This is a result of calling `apps.get_model('sites', 'Site')` ) 

I could try to do so inside the function but I am not sure if this will be a good practice.

Let me know what you think

joshblum: I don't think it's necessary to call `apps.get_model('sites', 'Site')`, can you just import it normally?

registration/models.py
joshblum: i think you should switch the lines since it actually is a bug in the original logic. the profile is deleted at L637, then L638 throws an exception and we try to delete the already deleted `profile` at L640

registration/models.py
joshblum: Won't you inherit the decorator though? So the child class should have `__unicode__` and `__str__` defined? (Sorry if this is what you meant and I just misunderstood)

registration/templates/registration/activation_complete_admin_pending.html
joshblum: what do you mean? If you use the example app to go through the registration flow just take some screenshots of what these pages look like, it's an easy way to catch small bugs on the templates in my opinion

registration/tests/enhanced_backend.py
joshblum: sounds good :)

test_app/settings.py
joshblum: yes i agree but the test app could be configured to use the `default` backend instead of yours so `ADMINS` does not need to be defined (except in the test settings)

registration/models.py
dicato: I would think you would be able to import it normally and then fetch the current site at runtime:

```python
from django.contrib.sites.models import Site
...
site = Site.objects.get_current()

File: registration/templates/registration/activation_complete_admin_pending.html (hunk 17 above)
Comment (sergafts) [inline registration/templates/registration/activation_complete_admin_pending.html]: Ah, I thought you wanted screenshots of the templates used create the new emails (towards admins, user)

Ok, the first new template is `activation_complete_admin_pending.html`
screenshot from 2016-06-21 10 18 19

And the second is 
screenshot from 2016-06-21 10 19 18

File: registration/models.py (hunk 14 above)
Comment (sergafts) [inline registration/models.py]: Hmm, I guess my class inherits from the decorated class so I should probably be fine. I will remove the decorator.

**Comment by sergafts** [comment]
@joshblum Hello again. I have given it a second run and implemented most of the ideas we discussed about. I have not done anything about the urls, since we have not reached to a conclusion. I am happy to implement it if we have a final decision.

Take a look and let me know.

File: registration/models.py [hunk 31]
```diff
@@ -31,6 +65,19 @@ class RegistrationManager(models.Manager):
 
     """
 
+    def _activate(self, profile, get_profile):
```
joshblum: add a short docstring for a description :)

registration/models.py
```diff
@@ -371,3 +408,267 @@ def send_activation_email(self, site, request=None):
...
+            self.send_admin_approve_email(profile.user, site)
+
+        # do not set ``User.is_active`` as True. This will be set
+        # when a site administrator approves this account.
+        profile.activated = True
+        profile.save()
+        if get_profile:
+            return profile
+        else:
+            return True
```
joshblum: should this return user? the docstring for `activate_user` doesn't expect a `True`, only `False` or a `User` (or `RegistrationProfile`) object

registration/models.py
joshblum: ^^ push this change

registration/models.py
```diff
@@ -371,3 +408,267 @@ def send_activation_email(self, site, request=None):
...
+            raise ImproperlyConfigured(
+                'Using the admin_approval registration backend'
+                ' requires at least one admin in settings.ADMINS')
+        admins = [admin[1] for admin in admins]
+        send_email(
+            admins, ctx_dict, admin_approve_email_subject,
+            admin_approve_email_body, admin_approve_email_html
+        )
+
+    def delete_expired_users(self):
```
joshblum: I thought we decided to remove this?

registration/models.py
joshblum: is the `__str__` method still needed? Did you try with the decorator on the super class and the `__str__` removed from the child

registration/templates/registration/activation_complete_admin_pending.html
joshblum: thanks!!

registration/tests/models.py
```diff
@@ -471,17 +474,324 @@ def test_resend_activation_email_expired_user(self):
...
+
+    """
+
+    user_info = {'username': 'alice',
+                 'password': 'swordfish',
+                 'email': 'alice@example.com'}
+
+    registration_profile = SupervisedRegistrationProfile
+
+    def setUp(self):
```
joshblum: use https://docs.djangoproject.com/en/1.9/topics/testing/tools/#django.test.override_settings instead of explict changes for `setUp` and `tearDown`

registration/tests/models.py
```diff
@@ -471,17 +474,324 @@ def test_resend_activation_email_expired_user(self):
...
+        new_user = self.registration_profile.objects.create_inactive_user(
+            site=Site.objects.get_current(), **self.user_info)
+        profile = self.registration_profile.objects.get(user=new_user)
+
+        user = self.registration_profile.objects.admin_approve_user(
+            profile.id, Site.objects.get_current())
+        self.failIf(isinstance(user, UserModel()))
+        self.assertEqual(user, False)
+
+    def test_activation_already_activated(self):
```
joshblum: remove?

registration/tests/models.py
```diff
@@ -471,17 +474,324 @@ def test_resend_activation_email_expired_user(self):
...
+        new_user = self.registration_profile.objects.create_inactive_user(
+            site=Site.objects.get_current(), **self.user_info)
+        profile = self.registration_profile.objects.get(user=new_user)
+        self.registration_profile.objects.activate_user(profile.activation_key)
+
+        profile = self.registration_profile.objects.get(user=new_user)
+        self.assertEqual(SupervisedRegistrationProfile.objects.activate_user(
+            profile.activation_key), False)
+
+    def test_activation_deactivated(self):
```
joshblum: remove?

registration/tests/models.py
```diff
@@ -471,17 +474,324 @@ def test_resend_activation_email_expired_user(self):
...
+        # Deactivate the new user.
+        new_user.is_active = False
+        new_user.save()
+
+        # Try to activate again and ensure False is returned.
+        failed = self.registration_profile.objects.activate_user(
+            profile.activation_key)
+        self.assertFalse(failed)
+
+    def test_activation_nonexistent_key(self):
```
joshblum: remove?

registration/tests/models.py
```diff
@@ -471,17 +474,324 @@ def test_resend_activation_email_expired_user(self):
...
+        associated with any account) fails.
+
+        """
+        # Due to the way activation keys are constructed during
+        # registration, this will never be a valid key.
+        invalid_key = hashlib.sha1(six.b('foo')).hexdigest()
+        self.failIf(SupervisedRegistrationProfile.objects.activate_user(
+            invalid_key))
+
+    def test_expired_user_deletion(self):
```
joshblum: remove?

registration/views.py
```diff
@@ -136,4 +135,31 @@ def render_form_submitted_template(self, form):
...
+            success_url = self.get_success_url(approved_user)
+            try:
+                to, args, kwargs = success_url
+            except ValueError:
+                return redirect(success_url)
+            else:
+                return redirect(to, *args, **kwargs)
+        return super(ApprovalView, self).get(request, *args, **kwargs)
+
+    def admin_approve(self, *args, **kwargs):
```
joshblum: rename to `approve` to be more generic

joshblum:
@sergafts sorry for the delay!! Left a few comments (and a few still need to be addressed) but it is looking good! Thanks for bearing with this rather slow process :)

registration/tests/models.py
sergafts: This is up for discussion. 
``RegistrationProfile.activate_user()`` includes the following:


    if profile.activated:
        # The User has already activated and is trying to activate
        # again. If the User is active, return the User. Else,
        # return False as the User has been deactivated by a site
        # administrator.
            if profile.user.is_active:
                return profile.user
            else:
                return False

When using the ``default`` backend, if a user tries to activate again **and** ``is_active`` is ``True`` this the user instance as a sign. This logic indicates that the user has activated his account succesfully. 

However, when using the ``admin_approval`` backend, if the user has completed activation succesfully ``is_active`` is ``False``. So this test will fail if it is inherited.

The issue is that when using different backends and by following the same procedure ``RegistrationProfile.activate_user()`` returns different results. 
I am not sure if you consider this an issue. Think about it and let me know. I think that this test, however needs to stay.

registration/tests/models.py
sergafts: If I override those settings, will they later be restored as they are in ``tearDown()`` ?

joshblum: yup it only lasts for the method/class that is overridden :)

registration/tests/models.py
joshblum: i think it's ok

registration/backends/admin_approval/urls.py
```diff
@@ -0,0 +1,77 @@
+"""
```
joshblum: let's just inherit from the `default` backend and add only the new urls needed

registration/tests/models.py
```diff
@@ -48,6 +50,7 @@ def tearDown(self):
         settings.REGISTRATION_DEFAULT_FROM_EMAIL = self.old_reg_email
         settings.REGISTRATION_EMAIL_HTML = self.old_email_html
         settings.DEFAULT_FROM_EMAIL = self.old_django_email
+        UserModel().objects.all().delete()
```
joshblum: why is this needed?

registration/tests/models.py
```diff
@@ -471,17 +474,332 @@ def test_resend_activation_email_expired_user(self):
...
+            ('Flea', 'admin2@iamaflea.com')
+        )
+
+    def tearDown(self):
+        settings.ACCOUNT_ACTIVATION_DAYS = self.old_activation
+        settings.REGISTRATION_DEFAULT_FROM_EMAIL = self.old_reg_email
+        settings.REGISTRATION_EMAIL_HTML = self.old_email_html
+        settings.DEFAULT_FROM_EMAIL = self.old_django_email
+
+    def test_valid_activation(self):
```
joshblum: remove?

registration/tests/models.py
```diff
@@ -471,17 +474,332 @@ def test_resend_activation_email_expired_user(self):
...
+                     .activate_user(profile.activation_key))
+
+        self.failUnless(isinstance(activated, UserModel()))
+        self.assertEqual(activated.id, new_user.id)
+        self.failIf(activated.is_active)
+
+        profile = self.registration_profile.objects.get(user=new_user)
+        self.assertTrue(profile.activated)
+
+    def test_valid_activation_with_profile(self):
```
joshblum: remove?

registration/tests/models.py
```diff
@@ -471,17 +474,332 @@ def test_resend_activation_email_expired_user(self):
...
+
+        self.failUnless(isinstance(activated_profile, self.registration_profile))
+        self.assertEqual(activated_profile.id, profile.id)
+        self.failUnless(activated_profile.activated)
+
+        new_user.refresh_from_db()
+        self.assertTrue(activated_profile.user.id, new_user.id)
+        self.assertFalse(new_user.is_active)
+
+    def test_resend_activation_email_activated_user(self):
```
joshblum: remove?

joshblum:
@sergafts just a few small nits and then I'll merge!

joshblum:
Also, @sergafts once complete can you squash your commits into a single one? Checkout http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git!

joshblum:
@sergafts Any updates?

sergafts:
@joshblum Hello. Sorry for the delay but I have been swamped lately. I hope I will have some time today to finish the process.

registration/backends/admin_approval/urls.py
sergafts: I am not sure whether this is possible after all. 

The urls in ``admin_approval/urls.py`` use different views than the ones in ``default/urls.py``. The difference is that the first ones use the new ``SupervisedRegistrationProfile``. I am not sure how to override this or if it is worth doing so. What do you think?

registration/tests/models.py
sergafts: This is different than ``RegistrationModelTests.test_valid_activation()`` as the latter has the following:

    self.failUnless(activated.is_active)

When using the default backend, after activation ``UserModel.is_active`` is set to ``True`` but when using the admin_approval backend it is set to ``False`` until account is approved.

registration/tests/models.py
sergafts: Same with ``test_valid_activation()`` above, ``new_user.is_active`` should be ``False``

registration/tests/models.py
sergafts: Same as two functions above. Also, when using the admin_approve backend after activation outbox will not be empty, instead it will have one email, the mail sent to admins.

sergafts:
I see Django1.10 builds are failing. Were these builds added now? I honestly can't recall if they existed before.

sergafts:
Also, as far as squashing is concerned, I have no problem doing so but I have an issue. My commits are on the master branch where additional commits have been added. If I rebase the commits that are between my commits will be squashed too right? Do you know of a way around this? (If the issue is not clear I can explain it better)

registration/tests/models.py
joshblum: ok cool let's leave as is then, can you add a comment noting the difference?

joshblum:
Django 1.10 support is new, if you need help to get the tests to pass I can try to take a look (also new to Django 1.10) @timgraham might have some helpful insights. As for squashing I'll handle it before merging :)

registration/backends/admin_approval/urls.py
joshblum: let's leave as is then!

registration/templates/registration/admin_approve_complete_email.txt
```diff
@@ -0,0 +1,13 @@
+{% load i18n %}
+{% blocktrans %}
+Your account is now approved. You can log in using the following link
+{% endblocktrans %}
+http://{{site.domain}}{% url 'django.contrib.auth.views.login' %}
```
timgraham: The test failures on 1.10 are because reversing by dotted path is removed. These templates should reverse by URL name (e.g. 'login') instead of by dotted path.

registration/views.py
```diff
@@ -136,4 +135,31 @@ def render_form_submitted_template(self, form):
...
+            except ValueError:
+                return redirect(success_url)
+            else:
+                return redirect(to, *args, **kwargs)
+        return super(ApprovalView, self).get(request, *args, **kwargs)
+
+    def approve(self, *args, **kwargs):
+        """
+        Implement admin-approval logic here.
+
```
timgraham: Remove the empty lines at the end of docstrings.

registration/templates/registration/admin_approve_complete_email.txt
joshblum: thanks!!

sergafts:
Not sure why tests continue to fail, specifically in Django 1.10. Would you mind taking a look too?

registration/tests/admin_approval_backend.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    Test the admin_approval registration backend.
+
+    Running these tests successfully will require two templates to be
+    created for the sending of activation emails; details on these
+    templates and their contexts may be found in the documentation for
+    the default backend.
+
+    """
+    urls = 'test_app.urls_admin_approval'
```
timgraham: Replace this with `@override_settings(ROOT_URLCONF=...)` as done in 80c1604c672a4cf382185103e5df9d2958563852.

joshblum:
Thanks @sergafts for the contribution!!

sergafts:
@joshblum My pleasure. I could also provide documentation if you wish.

joshblum:
That would be awesome :)

dicato:
@sergafts and @joshblum thanks for the hard work on this one!

In addition to the documentation, can either of you please update the changelog? Keeping it up to date makes cutting a release much easier. :smile:

sergafts:
@dicato Sure, I guess I can do that too, I will get to it after completing the documentation, I think it will be easier.

PR #189, opened 2016-06-10
Automatically compile locales from django.po to django.mo when building.

### LINKED ISSUE #188: Compile locales in setup.py
You can use this snippet in ``setup.py`` to compile locales while installing:

    class install_lib(_install_lib):
        def run(self):
            from django.core.management.commands.compilemessages \
                import compile_messages
            os.chdir('registration')
            compile_messages(sys.stderr)
            os.chdir("..")

allo-:
fixes #188


setup.py
```diff
@@ -17,6 +20,35 @@ def run_tests(self):
...
+
+class install_lib(_install_lib):
+    """
+        install_lib class extended to compile the locales
+        from the .po source files in .mo files.
+    """
+    def run(self):
+        _install_lib.run(self)
+        try:
+            # django before 1.7
```
joshblum: given #191 we can remove this

joshblum: @allo- can you update to remove the django 1.7 changes since we no longer support this version?

allo-:
If you really want this ... because it should not hurt on newer versions and people still using 1.7 could use it.

joshblum:
@allo- we explicitly dropped django 1.7 support so it would be better to fail loudly in my opinion

allo-:
Maybe there should be an explicit check? On the other hand ,the requirements.txt will produce a conflict for wrong django anyway. I added some comment on tested versions and how it works.

Not sure how stable the api for compile messages is. But as using the ``.handle()`` method corresponds to immitating the command line, it will probably work as long as command handling in django doesn't change.

setup.py
```diff
@@ -17,6 +20,26 @@ def run_tests(self):
...
+    """
+    def run(self):
+        _install_lib.run(self)
+        try:
+            # Use the management command to compile messages
+            # tested with django 1.7-1.9, 1.9 does not need the chdir anymore
+            from django.core.management.commands.compilemessages import Command
+            os.chdir('registration')
+            cmd = Command()
+            cmd.handle(verbosity=0, exclude=[], locale=[])
```
joshblum: can you use the `call_command` api? https://docs.djangoproject.com/en/1.9/ref/django-admin/#running-management-commands-from-your-code

allo-:
I guess thats the cleaner way. Need to look into it, when i have the time to test it.

joshblum:
@allo- any updates?

allo-:
Is there a reasonble option for testing the install in the way, that a install should create .mo files? This might be a useful unit test.

setup.py
```diff
@@ -17,6 +20,25 @@ def run_tests(self):
...
+    def run(self):
+        _install_lib.run(self)
+        try:
+            # Use the management command to compile messages
+            # tested with django 1.7-1.9, 1.9 does not need the chdir anymore
+            from django.core.management import call_command
+            os.chdir('registration')
+            call_command("compilemessages")
+            os.chdir('..')
+            return
```
joshblum: remove

setup.py
```diff
@@ -17,6 +20,25 @@ def run_tests(self):
...
         sys.exit(errno)
 
+
+class install_lib(_install_lib):
+    """
+        install_lib class extended to compile the locales
+        from the .po source files in .mo files.
+    """
+    def run(self):
+        _install_lib.run(self)
```
joshblum: `super()`

setup.py
```diff
@@ -17,6 +20,25 @@ def run_tests(self):
...
+        try:
+            # Use the management command to compile messages
+            # tested with django 1.7-1.9, 1.9 does not need the chdir anymore
```
joshblum: 1.8

setup.py
```diff
@@ -17,6 +20,25 @@ def run_tests(self):
...
+            from django.core.management import call_command
```
joshblum: why not import at the top of the file?

joshblum:
@allo- a few small nits, then lgtm. @dicato any ideas for testing or should we leave as is?

allo-:
- "remove": I do not understand what you want to remove
- "super": Yes, should be changed.
- "1.8"? I tested this part of the code from 1.7 to 1.9 and it ran without problems. Or what do you mean?
- "import not on top": This allows to fail graceful when the import fails (i.e. wrong django version). Before i even had a setup.py checked in, which works with older versions as well.

setup.py
joshblum: we no longer support version 1.7, so this comment should just read from 1.8-1.9

setup.py
joshblum: this line is not needed

setup.py
joshblum: we only support django version 1.8+, so it shouldn't be problem right?

allo-:
I find it kind of hard to follow the thread, when your answers appear above at the line notes.

- So which line to remove? You're quoting the whole class there.
- I would keep the comment, as it's definitely correct. Of course we could change it to 1.8-1.9 but what's the gain?
- "we only support django version 1.8+": Which line shouldn't be the problem? The conditional import?

joshblum:
sorry for the confusion, the lines I commented were the relevant ones. 

- the `return` at L38 is not needed
- The comment is fine, it should just read 1.8-1.9. We don't support 1.7 so there is no reason to mention it
- I think that the import can be moved to the top of the file since we only support 1.8+, we shouldn't have any import errors

dicato:
@joshblum I think the only way to test is to try a few supported locales and ensure the templates have the correct messages.

allo-:
removed the "tested with" completely, as it runs with the currently supported versions. This was only interesting, if a wide range of versions should be supported. The return was left over from before where the next code block tried the old method.

joshblum:
@allo- thanks for making these small changes, can you also change L31 to call `super()`? 

I tried running `python setup.py install` and got this error message: `django.core.management.base.CommandError: Can't find msgfmt. Make sure you have GNU gettext tools 0.15 or newer installed.` 

I resolved with [this SO post](http://stackoverflow.com/questions/35101850/cant-find-msguniq-make-sure-you-have-gnu-gettext-tools-0-15-or-newer-installed). Can you add a few lines to the readme that the `gettext` package is needed for manual setup?

allo-:
I wonder if you need the gettext binary or just python gettext packages. Can the CI test this?

allo-:
Probably there is a python gettext dependency in django anyway.

Another question is, if ``install_lib`` is the best install step for this. I had this from a SO answer, but it seems that the class has several other methods which can be used.

joshblum:
Rebasing would be great :) I don't have a strong opinion on the `install_lib` part, it seems to place the files in the correct location. What do you think?

allo-:
Err ... i used super the wrong way, but the other way it still fails, because ´´install_lib´´ is a old style class.

So either revert the commit and keep with the ``_install_lib.run(self)`` or try multiple inheritance from ``_install_lib`` and ``object`` (rather ugly, isn't it?)

allo-:
``install_lib`` isn't run i.e. for ``setup.py bdist_egg``. I think all binary packages should include prebuild ``.mo`` files.

joshblum:
Hmm let's just revert the commit then. What is the fix for `bdist_egg` then?

allo-:
Seems that neither ``bdist`` nor ``bdist_egg`` commands work as intended with ``install_lib``.

When i use ``build_py`` bdist works, but bdist_egg doesn't.

allo-:
The difference will be, if the target system of a binary distribution needs the ``gettext`` binary.

joshblum:
@allo- Thanks for the contribution and your patience during the review process!!

allo-:
Mh, i wanted to rebase it before a merge with less commits and a more readable commit message. 

If you like, push --force the revision before and i cleanup this branch later for a nicer merge.

And i am still not sure about the right command, but ``build_py`` will work until we find something better, which can be integrated then.

joshblum:
I squashed all of the commits in the end so it was just a single commit
with a combination of the other messages :)

allo-:
I would have reworded it, for example the one commit and it's exact revert aren't important in the log.

allo-:
If possible, i would like you to use
https://github.com/allo-/django-registration/commit/df4a36b22126736872e5a269d0514ec541917db5
I not only wrote a more concise commit message, but also fixed the real
name field of the commit (which just had the username in the commits you
squashed).

joshblum:
@allo- checkout https://github.com/macropin/django-registration/commit/a1b31a087027502c0c989dfcbf658c617cb28f3c i rewrote the commit message

allo-:
ah, okay.

joshblum:
all good?

allo-:
The name Metadata is still just the nick, but just let it be. No big deal for a few lines.

PR #197, opened 2016-06-23
Update copyright dates

SebastinSanty:


dicato:
Thanks! This made me realize we should probably have a discussion about the license and copyright given the original author is no longer involved. It would appear we lack the proper attribution to those who carried the project after the original author left. Filed #198

SebastinSanty:
Indeed I was also thinking of the same matter, seeing the authors have changed, but was afraid to change the names without proper consent. Instead chose to notify the authors, by atleast changing the dates.

Issue #213, opened 2016-09-09
create_new_activation_key is insecure


keattang:
It uses the python `random.random` which is not cryptographically secure and also unnecessarily uses the user's ID in the hash which makes it possible (although improbable) for the same key to be generated for a different user if their IDs overlap. For example, a user is created with an ID, then at some point deleted and a new user is created with the same ID.

I suggest using an algorithm that uses a secure random function to create an entirely random string or at least allow the user to specify their own algorithm.

Such an algorithm may look like this:

```
def create_new_activation_key(self, save=True):
        """
        Create a new activation key for the user
        """
        self.activation_key = hashlib.sha1(os.urandom(32)).hexdigest()
        if save:
            self.save()
        return self.activation_key
```

I am happy to make a PR if this is something you will consider merging?

dicato:
We would be happy to accept a pull request to address this issue!

Changing `random.random` to `os.urandom` is a good approach! I don't see any reason why this change would have backwards incompatible side effects, but please think through that as well. Thanks!

PR #214, opened 2016-09-12
Using a cryptographically secure method to generate activation keys

keattang:
I've used Django's `get_random_string` as this uses `random.SystemRandom()`/`os.urandom` and handles the case in which they are not available. You can see the source code here:
https://github.com/django/django/blob/stable/1.8.x/django/utils/crypto.py#L53

As far as I can tell there should be no backwards compatibility issues. I have created a test that should help prove this.

### DIFF
--- a/registration/models.py
+++ b/registration/models.py
@@ -2,8 +2,8 @@ from __future__ import unicode_literals
 
 import datetime
 import hashlib
-import random
 import re
+import string
 
 from django.conf import settings
 from django.contrib.sites.models import Site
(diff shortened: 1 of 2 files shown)


joshblum:
@keattang this looks great! thanks!!

joshblum:
closes #213

dicato:
@joshblum // @keattang can one of you please ensure this is recorded in the changelog?

Thanks!

keattang:
@dicato  See: https://github.com/macropin/django-registration/pull/215

Not sure why the pull request is showing the whole change again though...

keattang:
Any idea on an ETA for the next release?

PR #224, opened 2016-12-22
Fixes multilines in spanish (es) translation.

pchinea:
Adds missing space characters in multi-lines for spanish (es) translation.


dicato:
Thanks!

PR #229, opened 2017-01-24
Revert "Automatically compile locales from django.po to django.mo. (#…

### LINKED ISSUE #206: setup.py broke for new environments 23 days ago
https://github.com/macropin/django-registration/commit/a1b31a087027502c0c989dfcbf658c617cb28f3c

importing django into setup.py is a piss-poor idea that will break CI environments using this package, to speak nothing of new installs.

dicato:
…189)"

This reverts commit a1b31a087027502c0c989dfcbf658c617cb28f3c.

----

This adds the compiled files back to the repository and attempts to address the discussion in #206. I'm not very confident in this approach, but it's better than new installs being broken.

Thoughts?

Fix #206


joshblum:
Probably best to merge this (after lint fix) and do the new release

joshblum:
@dicato Can we land this to perform the release? Or did we want to try another approach?

dicato:
@joshblum I think so. I haven't had a moment to look at the flake warnings, but given this is a revert I'm not overly concerned.

joshblum:
The flake8 package was upgraded so the revert caused the lint. Just push a fix so we should be good to go :)

joshblum:
@dicato can you bump the pypi version or is that just @macropin ?

dicato:
Just @macropin unfortunately!

PR #235, opened 2017-02-15
[Add] check username case insensitive

pando85:


dicato:
I appreciate filing an issue and working on a pull request! Thank you for that!

Usernames are case sensitive in Django, correct? 

I'm not opposed to this change, as other packages like django-allauth have this same behavior. I'll leave some feedback on the actual code as well.

pando85:
I'm cleaning the code and the commits, please reopen.

pando85:
I appreciate any comments about this pull request.
Now, I'm going to open a new one to discuss doing a to_lowercase.

PR #236, opened 2017-02-16
Username case insensitive and username to lowercase

pando85:
I think if we do a username case insensitive is important to change user to lowercase for efficiency queries against database.


dicato:
@pando85 my comments here - https://github.com/macropin/django-registration/pull/235#issuecomment-280356805 - still apply to this new PR.

pando85:
@dicato  reopen #235 and discuss here the new propose to do an lower() before save.

joshblum:
@pando85 are you still working on this? Can you rebase onto master cleanly so we can review your proposed changes?

pando85:
Continue in #251

PR #237, opened 2017-02-20
[Add] check username case insensitive

pando85:
* [Add] check username case insensitive


joshblum:
closing (duplicated to #236)

PR #244, opened 2017-04-06
Make Site as optional

Surgo:
In Django 1.11 and when not use `Site` framework, following error occured.

```
RuntimeError: Model class django.contrib.sites.models.Site doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.
```

Use a shortcut.

### DIFF
--- a/registration/models.py
+++ b/registration/models.py
@@ -6,7 +6,7 @@ import re
 import string
 
 from django.conf import settings
-from django.contrib.sites.models import Site
+from django.contrib.sites.shortcuts import get_current_site
 from django.core.exceptions import ImproperlyConfigured
 from django.core.exceptions import ObjectDoesNotExist
 from django.core.mail import EmailMultiAlternatives
@@ -422,7 +422,7 @@ class SupervisedRegistrationManager(RegistrationManager):
         """
 
         if not profile.user.is_active and not profile.activated:
-            site = Site.objects.get_current()
+            site = get_current_site()
             self.send_admin_approve_email(profile.user, site)
 
         # do not set ``User.is_active`` as True. This will be set
--- a/registration/models.py
+++ b/registration/models.py
@@ -6,7 +6,6 @@ import re
 import string
 
 from django.conf import settings
-from django.contrib.sites.shortcuts import get_current_site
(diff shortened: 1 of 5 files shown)


joshblum:
@Surgo thanks!! Will merge once tests pass :)

registration/models.py
```diff
@@ -422,7 +422,7 @@ def _activate(self, profile, get_profile):
         """
 
         if not profile.user.is_active and not profile.activated:
-            site = Site.objects.get_current()
+            site = get_current_site()
```
joshblum: you have to pass the `request` in https://docs.djangoproject.com/en/1.8/ref/contrib/sites/#django.contrib.sites.shortcuts.get_current_site :)

Surgo: Oooooops. I'll fix it :(

Surgo: OMG... May be it's heavy...

joshblum: Does the fix not work?

Surgo: If `SupervisedRegistrationManager._activate` interface is changed (like `_activate(self, site, profile, get_profile)`), have to change `RegistrationManager._activate` and `RegistrationManager.activate_user` as same as `SupervisedRegistrationManager`.

Its not have `requests` or `site`.

joshblum: can you implement this fix?

Surgo: Can I try to change `RegistrationManager.activate_user` interface to `activate_user(self, site, activation_key, get_profile=False)` like `RegistrationManager.create_inactive_user(self, site, new_user=None, send_email=True, request=None, profile_info={}, **user_info)`

Surgo: I'll try tomorrow :)

joshblum: I think that https://github.com/macropin/django-registration/blob/master/registration/admin.py#L37 can also be changed to use the `get_current_site` shortcut

Surgo: Thank you. I will fix it too:)

Surgo:
@joshblum Add `site` to `activate_user`'s arg :) and remove `Site`.

joshblum:
@Surgo huge thanks for putting this together so quickly!!

PR #249, opened 2017-04-08
Add script for releasing to pypi

joshblum:


dicato:
:boom: Should make releasing more straightforward.

PR #251, opened 2017-04-10
Username case insensitive and username to lowercase 

pando85:
Following #236.


pando85:
Why didn't you merge it yet? What do you need?

dicato:
@pando85, there are a couple issues with this pull request that need to be addressed before it can be reviewed and merged accordingly.

* tests are failing
* it needs to be rebased cleanly on against master
* it should not touch translations in the same pr

joshblum:
@pando85 any update on this?

registration/forms.py
```diff
@@ -39,6 +40,13 @@ class Meta:
         model = User
         fields = (UsernameField(), "email")
 
+    def clean_username(self):
+        username = self.cleaned_data.get('username')
```
joshblum: `username = self.cleaned_data.get('username', '')` since we call `lower()` in the line below

registration/forms.py
```diff
@@ -13,6 +13,7 @@
 from django import forms
 from django.utils.translation import ugettext_lazy as _
 from django.contrib.auth.forms import UserCreationForm
+from django.forms import ValidationError
```
joshblum: i know this wasn't your change but could you alphabetize the imports?

registration/forms.py
```diff
@@ -39,6 +40,13 @@ class Meta:
         model = User
         fields = (UsernameField(), "email")
 
+    def clean_username(self):
+        username = self.cleaned_data.get('username')
+        if User.objects.filter(username=username.lower()).exists():
+            raise ValidationError(_('A user with that username already exists.'))
```
joshblum: can you add a test for the new functionality?

registration/forms.py
joshblum: I would prefer you just call `lower()` here instead of in the conditional and on the return

registration/forms.py
```diff
@@ -39,6 +40,13 @@ class Meta:
         model = User
         fields = (UsernameField(), "email")
 
+    def clean_username(self):
+        username = self.cleaned_data.get('username').lower()
```
joshblum: this should return an empty string by default instead of `None`:

`username = self.cleaned_data.get('username', '').lower()`

https://docs.python.org/3/library/stdtypes.html#dict.get

pando85: Ah, ok, sorry. I'm trying to pass tests. I will change that too.

joshblum: no problem! just wanted to make sure you didn't miss it :)

pando85:
What kind of test cloud  be missing here?

registration/forms.py
```diff
@@ -40,6 +40,20 @@ class Meta:
...
 
+class RegistrationFormUsernameLowercase(RegistrationForm):
+    """
+    A subclass of :class:`RegistrationForm` which enforces unique case insensitive
+    usernames, make all usernames to lower case.
+
+    """
+    def clean_username(self):
+        username = self.cleaned_data.get('username', '').lower()
+        if User.objects.filter(email=username).exists():
```
joshblum: `UsernameField()=username`? why did you put `email=username`?

joshblum:
@pando85 looking good one final change and it should be good to merge! Thanks for pushing so hard on this! :)

registration/forms.py
pando85: Fixed. Terrible fail, sorry.

registration/forms.py
```diff
@@ -40,6 +40,20 @@ class Meta:
...
 
+class RegistrationFormUsernameLowercase(RegistrationForm):
+    """
+    A subclass of :class:`RegistrationForm` which enforces unique case insensitive
+    usernames, make all usernames to lower case.
+
+    """
+    def clean_username(self):
+        username = self.cleaned_data.get('username', '').lower()
+        if User.objects.filter(username=username).exists():
```
joshblum: use `UsernameField()=username` since the username field is customizable. otherwise the code is good to go!!

pando85: I don't know how to do that. Thanks for all your help!

pando85: Ok, I found it in stackoverflow: http://stackoverflow.com/questions/4720079/django-query-filter-with-variable-column.
It could be finished.

registration/forms.py
```diff
@@ -40,6 +40,20 @@ class Meta:
...
 
+class RegistrationFormUsernameLowercase(RegistrationForm):
+    """
+    A subclass of :class:`RegistrationForm` which enforces unique case insensitive
+    usernames, make all usernames to lower case.
+
+    """
+    def clean_username(self):
+        username = self.cleaned_data.get('username', '').lower()
+        if User.objects.filter(**{ UsernameField(): username}).exists():
```
joshblum: linting fails :/
`./registration/forms.py:51:35: E201 whitespace after '{'`

pando85:
Finally?
Can I overwrite my registration form with this parameter at settings: `REGISTRATION_FORM = "registration.forms.RegistrationFormUsernameLowercase"`?
Thanks for all!

joshblum:
@pando85 yes that will work for override!

joshblum:
(as soon as we release version 1.6 #255) you can `pip install --upgrade django-registration-redux` to get the latest version with your code!

PR #259, opened 2017-05-16
german translations completed

zypro:


dicato:
@zypro thank you! Merging!

PR #262, opened 2017-05-30
Redirect Logged Users

sammarxz:
If the user is already logged in, why show the registration page?

### DIFF
--- a/registration/tests/default_backend.py
+++ b/registration/tests/default_backend.py
@@ -1,6 +1,7 @@
 import datetime
 
 from django.conf import settings
+from django.contrib.auth.models import AnonymousUser
 from django.core import mail
 from django.core.urlresolvers import reverse
 from django.test import TestCase
@@ -129,11 +130,13 @@ class DefaultBackendViewTests(TestCase):
 
         request_factory = RequestFactory()
         view = RegistrationNoEmailView.as_view()
-        view(request_factory.post('/', data={
+        request = request_factory.post('/', data={
             'username': 'bob',
             'email': 'bob@example.com',
             'password1': 'secret',
-            'password2': 'secret'}))
+            'password2': 'secret'})
+        request.user = AnonymousUser()
+        view(request)
 
         UserModel().objects.get(username='bob')
         # A registration profile was created, and no activation email was sent.
--- a/registration/tests/simple_backend.py
+++ b/registration/tests/simple_backend.py
@@ -72,7 +72,7 @@ class SimpleBackendViewTests(TestCase):
         self.failUnless(new_user.is_active)
 
         # New user must be logged in.
-        resp = self.client.get(reverse('registration_register'))
+        resp = self.client.get(reverse('registration_register'), follow=True)
         self.failUnless(resp.context['user'].is_authenticated())
 
     def test_registration_failure(self):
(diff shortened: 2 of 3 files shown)


joshblum:
@sammarxz you should see why the tests are failing. Overall it seems reasonable, I don't have a strong opinion. @dicato is there any reason we wouldn't want this functionality?

dicato:
I agree. I do this in my projects for registration and login urls.

sammarxz:
I think it failed because the LOGIN_REDIRECT_URL is None (the user needs to set in the project settings)

dicato:
@sammarxz that's right. Thanks for making me think about this a little more.

I think we could copy the design pattern from django-allauth; you can see this [configuration docs here](http://django-allauth.readthedocs.io/en/latest/configuration.html#configuration).

Essentially, we should a settings specific to django-registration-redux, named `ACCOUNT_AUTHENTICATED_REGISTRATION_REDIRECTS`, that defaults to `True`, and conditionally redirects to `LOGIN_REDIRECT_URL`. 

If `LOGIN_REDIRECT_URL` is `None`, an exception should likely be thrown with a clear message to the author.

sammarxz:
@dicato all the tests failed, I could not understand .-.

ioparaskev:
@sammarxz in order for the test to pass you need to make some changes to 2 tests cases. One is related to passing a request user (since the request factory does not provide one) and the other is to follow redirects (since the new code redirects instead of an immediate respone. Below you can see the diff needed to be applied to your PR in order for the tests to pass:

```
diff --git a/registration/tests/default_backend.py b/registration/tests/default_backend.py
index f7eec35..83e56e8 100644
--- a/registration/tests/default_backend.py
+++ b/registration/tests/default_backend.py
@@ -6,6 +6,7 @@ from django.core.urlresolvers import reverse
 from django.test import TestCase
 from django.test.client import RequestFactory
 from django.test.utils import override_settings
+from django.contrib.auth.models import AnonymousUser

 from registration.forms import RegistrationForm
 from registration.backends.default.views import RegistrationView
@@ -129,11 +130,13 @@ class DefaultBackendViewTests(TestCase):

         request_factory = RequestFactory()
         view = RegistrationNoEmailView.as_view()
-        view(request_factory.post('/', data={
+        request = request_factory.post('/', data={
             'username': 'bob',
             'email': 'bob@example.com',
             'password1': 'secret',
-            'password2': 'secret'}))
+            'password2': 'secret'})
+        request.user = AnonymousUser()
+        view(request)

         UserModel().objects.get(username='bob')
         # A registration profile was created, and no activation email was sent.
diff --git a/registration/tests/simple_backend.py b/registration/tests/simple_backend.py
index 4777de9..7528c6d 100644
--- a/registration/tests/simple_backend.py
+++ b/registration/tests/simple_backend.py
@@ -72,7 +72,7 @@ class SimpleBackendViewTests(TestCase):
         self.failUnless(new_user.is_active)

         # New user must be logged in.
-        resp = self.client.get(reverse('registration_register'))
+        resp = self.client.get(reverse('registration_register'), follow=True)
         self.failUnless(resp.context['user'].is_authenticated())

     def test_registration_failure(self):
```

joshblum:
@sammarxz I've applied the changes that @ioparaskev suggested (thanks!) so we can merge this PR in for the next release. Thank you both for your help!!

Issue #261, opened 2017-05-27
Migrate failed when apply "0003_migrate_activatedstatus" on MySQL


Surgo:
`1146, "Table 'default.registration_registrationprofile' doesn't exist` is occured.

Why table name is `default.registration_registrationprofile`?

```
  Applying registration.0001_initial...DEBUG	2017-05-27 11:55:27,655	CREATE TABLE `registration_registrationprofile` (`id` integer AUTO_INCREMENT NOT NULL PRIMARY KEY, `activation_key` varchar(40) NOT NULL, `user_id` varchar(8) NOT NULL UNIQUE) ROW_FORMAT=DYNAMIC; (params None)
DEBUG	2017-05-27 11:55:27,707	ALTER TABLE `registration_registrationprofile` ADD CONSTRAINT `registration_registrationpr_user_id_5fcbf725_fk_accounts_user_id` FOREIGN KEY (`user_id`) REFERENCES `accounts_user` (`id`); (params [])
 OK (0.311s)
  Applying registration.0002_registrationprofile_activated...DEBUG	2017-05-27 11:55:28,059	ALTER TABLE `registration_registrationprofile` ADD COLUMN `activated` bool DEFAULT %s NOT NULL; (params [False])
DEBUG	2017-05-27 11:55:28,144	ALTER TABLE `registration_registrationprofile` ALTER COLUMN `activated` DROP DEFAULT; (params [])
 OK (0.304s)
  Applying registration.0003_migrate_activatedstatus...Traceback (most recent call last):
  File "/home/travis/build/test/webapp/.tox/python35-django110/lib/python3.5/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/home/travis/build/test/webapp/.tox/python35-django110/lib/python3.5/site-packages/django/db/backends/mysql/base.py", line 110, in execute
    return self.cursor.execute(query, args)
  File "/home/travis/build/test/webapp/.tox/python35-django110/lib/python3.5/site-packages/pymysql/cursors.py", line 166, in execute
    result = self._query(query)
  File "/home/travis/build/test/webapp/.tox/python35-django110/lib/python3.5/site-packages/pymysql/cursors.py", line 322, in _query
    conn.query(q)
  File "/home/travis/build/test/webapp/.tox/python35-django110/lib/python3.5/site-packages/pymysql/connections.py", line 856, in query
    self._affected_rows = self._read_query_result(unbuffered=unbuffered)
[...]

PR #264, opened 2017-05-31
Use current database connection's alias for testing. Fixes: #261

Surgo:
Fixes: #261

In [RunPython](https://docs.djangoproject.com/en/1.11/ref/migration-operations/#django.db.migrations.operations.RunPython), should specify current connection for testing.

You can see a example in above django docs link.

Because use alias in testing. 
May be same problems are occurred when use multiple databases.


joshblum:
lgtm thanks @Surgo

Issue #266, opened 2017-06-07
Use Django 1.11 class-based views when possible


joshblum:
The function based views are deprecated in Django 1.11 and fixes a security bug with potentially leaking the password reset token with the referrer header. https://docs.djangoproject.com/en/1.11/releases/1.11/#django-contrib-auth

joshblum:
closed by #268

PR #268, opened 2017-06-07
Prevent leaking password reset token through Referrer header

joshblum:
Addresses #266, preventing the leaking of password reset token through the `Referrer` header.

For Django 1.11+, we fix by using the (newer class based views)[https://docs.djangoproject.com/en/1.11/releases/1.11/#django-contrib-auth].

For versions of Django below 1.11 we add a `meta` block to the template add at the meta tag `<meta name="referrer" content="never">`. In addition, we add `rel="noreferrer"` to the password reset email.


joshblum:
@macropin @dicato I tried to go through a list of similar apps that might have issues to alert them: https://djangopackages.org/grids/g/registration/ 
If either of you know of any other packages that may be affected, let me know and i'll be happy to open something up :)

registration/auth_urls.py
```diff
@@ -27,35 +27,54 @@
 from django.core.urlresolvers import reverse_lazy
 from django.contrib.auth import views as auth_views
 
+# Attempt to use the auth class based views if available.
+try:
+    login_view = auth_views.LoginView.as_view()
```
joshblum: need to pass kwargs to `as_view` to override defaults

joshblum: done

dicato:
@joshblum sorry for the delay. Looks good to me!

Do you want to fix the use of "referrer" to "referer"? Probably good to be consistent with the [original misspelling](https://en.wikipedia.org/wiki/HTTP_referer).

joshblum:
@dicato great! fixed the misspellings and will merge/pypy bump after tests :)

PR #271, opened 2017-07-17
Fixed url in quickstart

RatanShreshtha:


joshblum:
@RatanShreshtha thanks for the PR! Can you quickly verify if the same mistake occurs elsewhere in the docs? Will merge after confirmation :)

PR #279, opened 2017-08-30
Add REGISTRATION_ADMINS mail setting for approval

ioparaskev:
Checks for REGISTRATION_ADMINS defined in settings to use for the
approval email instead of ADMINS. Triggers warning (about using ADMINS)
if setting is not found

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,9 @@ django-registration-redux changelog
 
 Version 1.8, TBD
 ----------------
+* Feature: Added `REGISTRATION_ADMINS` setting to support different admins (than those in `ADMINS`) in charge of registration.
+Falls back to `ADMINS` in case this is not set
+* Documentation: Added documentation regarding admin approval registration
 
 Version 1.7, 18 July, 2017
 ----------------
--- a/docs/admin-approval-backend.rst
+++ b/docs/admin-approval-backend.rst
@@ -1,0 +1,40 @@
+.. _admin-approval-backend:
+.. module:: registration.backends.admin_approval
+
+The admin approval backend
+==========================
+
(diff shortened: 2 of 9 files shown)


ioparaskev:
Checking about documenting this change, I noticed there's no documentation about which mail is used for the approval. I wasn't sure if this was the right point to document it, so I didn't make any changes.

joshblum:
@ioparaskev these changes look great! Would you be able to add a doc page about the `admin_approval` backend? Here are some example of docs for the [`default`](https://github.com/macropin/django-registration/blob/master/docs/default-backend.rst) or [`simple`](https://github.com/macropin/django-registration/blob/master/docs/simple-backend.rst). I know it's a lot to ask, but having it documented would be a great help. @sergafts and I missed added them when working on the initial implementation.

joshblum:
Additionally can you add a line to the CHANGELOG noting the change?

ioparaskev:
ok i'll make the changes in this PR as well

ioparaskev:
@joshblum I think it's documented enough now. From what I saw travis had a github related issue in a python 3.3 build, so my guess is that re-triggering the build will pass smoothly. Let me know if anything else is needed

ioparaskev:
@joshblum It seems there is indeed a python 3.3 issue. I'll try to set up a tox env and fix this asap

ioparaskev:
@joshblum It seems this is a travis/travis config issue.
If you check [here](https://travis-ci.org/ioparaskev/django-registration/jobs/272494216) you'll see that the specific travis env also fails for the current master branch.. Seems like an issue with the discovery of tests. I suggest we see this in another PR (I also have a tox.ini to share to help debug such cases which I'll add in another PR)

PR #283, opened 2017-09-08
Add possible settings for activation emails

pae4557:
The django settings file may dictate where registration.models.RegistrationProfile.send_activation_email gets templates from. These are the email templates that are sent to a user when they sign up in order to activate their account, and to make custom emails the template source must be changed.


joshblum:
@pae4557 thanks for the contribution!

PR #290, opened 2017-10-02
Fix travis builds

joshblum:
Travis builds non-deterministically fail for python 3.3 due to different versions of the `pytest` library being installed. We now force this package to be `>3.2.1` for builds to pass. 

@ioparaskev this fixes the issue you were going to look at!


ioparaskev:
@joshblum I didn't have time to check this earlier. I'm glad you found a working solution ^^

Issue #26, opened 2015-02-03
not being able to define per site email address


sircco:
Since it is already possible to use site for email notifications it would be great if it could be possible
to override settings.DEFAULT_FROM_ADDRESS  from function

macropin:
Somewhat related to #54

joshblum:
@sircco does the fix in https://github.com/macropin/django-registration/pull/54 solve your problem?

sircco:
looking at code, i don't think that this does it unless i can  override this line on per case basis.
`  from_email = getattr(settings, 'REGISTRATION_DEFAULT_FROM_EMAIL', settings.DEFAULT_FROM_EMAIL)`

Our app is whitelabel, meaning that more than one domain can use it, register its own customers (and send registration emails)

dicato:
@sircco the use case makes sense 👍 

If you'd like to implement we are happy to review.

joshblum:
strawman could be to allow `REGISTRATION_DEFAULT_FROM_EMAIL` to be a python dotted `path.to.function` which would point a function that take `site` as an argument and does custom logic to return the email? Thoughts?

```
# path/to.py
def function(site):
   return 'example@example.com'
```

joshblum:
(probably cleaner to have a separate settings variable `REGISTRATION_FROM_EMAIL_FUNCTION` but same idea)

sircco:
How about something like this, overriding default if variable not provided
during instantiation.
I didnt look too much at the code recently but here is some dummy code to
explain my idea.

class Registration(Object):
   def __init___(self, mail_from=None, .... ):
       self.mail_from, = mail_from or
settings.REGISTRATION_DEFAULT_FROM_EMAIL

dicato:
I think it's acceptable to use static configuration for this as well. You can use a dictionary to map emails to sites, just as an example.

joshblum:
@sircco would a static configuration meet your use case?

sircco:
It will work for now. If we get so much traffic on our whitelabel site that we can't manage static setting we will think of something and submit code for review.

joshblum:
Great. Do you want to take a first stab at implementing this?

joshblum:
@sircco take a look at #294 and let me know if this address your use case!

sircco:
should work fine!

joshblum:
Awesome! I'll leave this open until @294 is merged in

joshblum:
closed by #294

PR #294, opened 2017-10-04
Configure from email from Site object

joshblum:
Addresses #26

### DIFF
--- a/registration/models.py
+++ b/registration/models.py
@@ -27,6 +27,13 @@ SHA1_RE = re.compile('^[a-f0-9]{40}$')
 
 
 def get_from_email(site=None):
+    """
+    Return the email address by which mail is sent.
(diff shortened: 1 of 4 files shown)


registration/models.py
```diff
@@ -25,6 +26,22 @@
 SHA1_RE = re.compile('^[a-f0-9]{40}$')
 
 
+def get_from_email(site=None):
+    if getattr(settings, 'REGISTRATION_USE_SITE_EMAIL', False):
```
dicato: We should probably add a docstring here.

dicato:
This looks straightforward. Are there any manual testing steps?

registration/models.py
joshblum: done

joshblum:
No manual testing tests, just the added unit tests

Issue #303, opened 2017-10-17
Missing registration_complete translation.


jorrete:
Hi,
The registration_complete translation string is missing in po file.
Thanks!

joshblum:
@jorrete closing this for now since all relevant strings should appear in the `.po` files, we just need the valid translations added :)

PR #304, opened 2017-10-17
Make messages to update all .po files

joshblum:
@jorrete I reran `django-admin makemessages -a` and ` django-admin compilemessages` to gather any new strings for translation. Does this address your issue in #303? I'm not very familiar with working with translations so let me know if this addresses the issue to allow someone to fill in the appropriate translation. @dicato let me know what you think :)


dicato:
@jorrete - was there a specific string that was missing for a given translation? 

@joshblum I believe your steps to make and compile translations is correct.

joshblum:
@dicato thanks! Will merge for now and we can update the missing translations as needed.

Issue #233, opened 2017-02-09
Log accounts being deleted in routine cleanup task


felixSchl:
We just lost about 10 valid accounts and are in the process of recovering them, unfortunately at this point I am guessing that it must have been django-registration's clean up task but there's no way of knowing for sure, the data is just gone. If it logged the accounts to be deleted before actual deletion, I could go back to be cron job's log file and check what happened and when.

dicato:
Are you calling the `cleanupregistration` management command via cron?

Looking at the [tests](https://github.com/macropin/django-registration/blob/37a881cdf11eb38e96a5b830201e238f96151936/registration/tests/models.py#L349) for the [delete_expired_users](https://github.com/macropin/django-registration/blob/37a881cdf11eb38e96a5b830201e238f96151936/registration/models.py#L201) method, it certainly could be improved.

Have you tried to recreate this issue? Can you?

felixSchl:
Yes, calling it via cron and piping the output to a file, but there's no output.

I'd say the core issue here is the lack of overview of what's happening. If it was django registration, the only thing I can think of are timezone issues or something like that. I will spend time looking into this further after I recovered the accounts properly.

dicato:
@felixSchl I can appreciate the frustration here and desire to figure out the problem. The lack of comprehensive tests for this is disappointing and something that should clearly be fixed.

felixSchl:
I just noticed something else, i think [this](https://github.com/macropin/django-registration/blob/master/registration/models.py#L243) should read:

```
 if profile.activation_key_expired() and not profile.activated:
```

dicato:
@felixSchl you are probably right regarding that condition, since it's clearer. However the logic currently uses `if not user.is_active:` so it doesn't seem like a bug.

felixSchl:
But `user.is_active` is something that could change irrespective of initial successful registration, right? For example when temporarily suspending an account.

dicato:
Great point! Yes, I can absolutely see that use case causing inactive accounts being deleted erroneously.

We should:

1) fix the logic to use `if profile.activation_key_expired() and not profile.activated:`
2) add/improve the unit tests for this
3) add logging to the management command that uses this code

joshblum:
@felixSchl are you able to submit a patch to fix this?

PR #308, opened 2017-10-30
Add logging to cleanup method and use profile.activated

joshblum:
Addresses #233. @felixSchl sorry this so long to get around to. let me know if this addresses your issue! @dicato wasn't sure what you had in mind for beefing up the unit tests but if you have any suggestions happy to throw them in

### DIFF
--- a/registration/models.py
+++ b/registration/models.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 
 import datetime
 import hashlib
+import logging
 import re
 import string
 import warnings
@@ -24,6 +25,7 @@ from django.utils.translation import ugettext_lazy as _
 from .users import UserModel
 from .users import UserModelString
 
+logger = logging.getLogger(__name__)
 SHA1_RE = re.compile('^[a-f0-9]{40}$')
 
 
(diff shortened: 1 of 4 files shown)


dicato:
@joshblum regarding the tests, if I remember correctly you have three attributes you are performing logic on:

* whether the profile is activated
* whether the user is active
* and, to some extent, whether the user exists

The unit tests did not cover all these situations.

registration/models.py
```diff
@@ -265,12 +266,13 @@ def delete_expired_users(self):
         """
         for profile in self.all():
             try:
-                if profile.activation_key_expired():
+                if profile.activation_key_expired() and not profile.activated:
                     user = profile.user
-                    if not user.is_active:
-                        profile.delete()
-                        user.delete()
+                    logger.warning('Deleting expired Registration profile {} and user {}.'.format(profile, user))
```
dicato: I don't believe `logging` is used anywhere else in this project; how will this impact users of this package? What happens if they don't have logging setup?

joshblum: It's not used elsewhere and I haven't noticed in a few other third party django apps I checked. But is there an alternative? We can suggest a default logging configuration in the docs and mention that it's necessary for the cron job unless there is a different approach we can take.

joshblum:
Awesome, added a test for an active and expired user. Let me know your opinion on adding a logging config to the docs or some other approach!

registration/models.py
dicato: I poked around a bit and convinced myself this should be good. :smile:

Other packages follow the same pattern.

joshblum: cool i'll merge in then!

Issue #240, opened 2017-03-17
Default ActivationView doesn't check whether user is already activated


delano:
In the [default ActivationView](https://github.com/macropin/django-registration/blob/v1.4/registration/backends/default/views.py#L119-L138), the call to `RegistrationProfile.activate_user` returns the `User` object when the user is first activated but also if the user is already activated. So in the view, there's no way for it to know whether it's the initial activation. 

```python
class ActivationView(BaseActivationView):
    def activate(self, *args, **kwargs):
        activation_key = kwargs.get('activation_key', '')
        activated_user = (RegistrationProfile.objects
                          .activate_user(activation_key))
        if activated_user:
            signals.user_activated.send(sender=self.__class__,
                                        user=activated_user,
                                        request=self.request)
        return activated_user
```

This leads to two issues:
* The `user_activated` signal is called each time a user clicks an activation link. Say for example, if this signal is used to send a notification, that notification will get sent every time.
* If `REGISTRATION_AUTO_LOGIN` is set to `True`, the activation link can be used as a login link indefinitely. 

Note: [this is an issue in the original `django-registration` as well](https://github.com/ubernostrum/django-registration/blob/2.2/registration/views.py#L93-L113)

dicato:
@delano thanks for filing this! Are you interested in trying to improve this?

I'll take a look through the code to refresh my memory before making any suggestions.

delano:
@dicato Yeah I can push a fix, no problem. Maybe tomorrow but within the next week for sure.

joshblum:
@delano are you still interested in working on this?

delano:
Yeah sorry for the delay!

delano:
@joshblum There are a couple options for this but nothing that doesn't change current expected functionality or isn't hacky. There is one option that isn't hacky but it requires a second hit to the DB. 

* Have `RegistrationManager.activate_user` return a tuple (e.g. `(profile_user, True)`). This would be similar to Django's `get_or_create` method but would require updating every call this this method (and would be a breaking change for end users that have custom registration views).
* Call the signal from within `RegistrationManager.activate_user` when `profile.activated` is false. Has side effects since this is called from places other than the view (and would need to pass is the `request` object).
* In `ActivationView.activate`, call the DB before `RegistrationManager.activate_user` to see if the profile was already activated.
* Delete the `RegistrationProfile` record when the user is activated or add `activated=False` to the query in `RegistrationManager.activate_user`. These will have the same side effects. 

[...]

PR #309, opened 2017-10-30
Change return signature of Registration_profile.activate_user to return a tuple

joshblum:
Addresses #240. @delano, went with the change of signature of ``RegistrationProfile.activate_user``. Would love any comments from you and @dicato before merging.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,10 @@ django-registration-redux changelog
 
 Version 1.9, TBD
 ----------------
+* BACKWARDS-INCOMPATIBLE CHANGE: Change return signature of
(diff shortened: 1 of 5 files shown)


CHANGELOG
```diff
@@ -4,6 +4,10 @@ django-registration-redux changelog
 
 Version 1.9, TBD
 ----------------
+* BACKWARDS-INCOMPATIBLE CHANGE: Change return signature of
+``RegistrationProfileManager.activate_user``. A tuple of (``User``,
+``activated(bool)``) is now returned.
```
dicato: It's probably clearer to say "A tuple containing the ``User`` instance and a boolean of whether or not said user was activated".

PR #326, opened 2018-02-16
Specified app_label to prevent issues with app_label when inheriting …

msert29:
Using the Registration package prompts app_label issue when migrating from Django 1.8 to 1.11. The specific reason for this was that because of a custom user model which inherited from registration model. Hence this fix is to prevent such errors.

### DIFF
--- a/registration/models.py
+++ b/registration/models.py
@@ -311,6 +311,7 @@ class RegistrationProfile(models.Model):
     class Meta:
         verbose_name = _('registration profile')
         verbose_name_plural = _('registration profiles')
+        app_label = 'registration'
 
     def __str__(self):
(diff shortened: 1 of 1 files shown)


joshblum:
@msert29 do you have you the traceback you get when encountering the error?

registration/models.py
```diff
@@ -628,6 +629,9 @@ class SupervisedRegistrationProfile(RegistrationProfile):
     # in admin approval
     objects = SupervisedRegistrationManager()
 
+    class Meta:
```
joshblum: Is this necessary? `SupervisedRegistrationProfile` inherits from `RegistrationProfile`

PR #333, opened 2018-04-11
Minor changes in Spanish translation

jantoniomartin:
After the last commit, I have made some minor changes in the Spanish translations. Some strings were missing and others were not correct. Also I have changed some strings to address the user formally (there is no difference in English, but there is in Spanish).


joshblum:
@jantoniomartin thanks!! i'll release an updated version this evening

PR #335, opened 2018-04-30
Changed path reference from 'login' to 'auth_login

roderik333:
This caused a trace-back when using the default templates.


joshblum:
hey @roderik333 i wasn't able to reproduce the error using the `test_app` -- can you add the traceback here?

PR #344, opened 2018-10-27
 Fix cleanupregistration, skip if user.is_active

joshblum:


joshblum:
`raise RuntimeError("Python 3.4 or later is required")`

PR #345, opened 2018-10-28
Admins getter

GasparPizarro:
The `REGISTRATION_ADMINS` can be a list of `(name, email)` pairs as always, or can be the import path of a callable that returns a list like that.


ZeroCoolHacker:
Nice work

PR #351, opened 2018-12-27
pypi: add source packaging too

marcaurele:
The source package is needed when installing packages with the `--no-binary=:all:` parameter (to `pip`).


marcaurele:
Is there a specific reason to not upload the source package anymore?

PR #355, opened 2019-02-22
Send email on commit

adibo:
I propose to use `on_commit` for sending out activation emails.

This could be useful when e.g. sending out activation emails are delegated to a celery task. Potential race condition exists, when a user object has yet not been committed to a db while a task broker already attempts to execute email sending. 
A similar problem is described here: https://dev.to/k4ml/django-fixing-race-condition-when-queuing-with-oncommit-hook-7ae

Since `django-registration` relies on Django 1.11+, we could safely use `on_commit` to avoid such situations, as sugested by the post.

The only caveat is that (some) tests would need to use `TransactionTestCase`. 
All the necessary changes are proposed in this PR.

### DIFF
--- a/test_app/settings_test.py
+++ b/test_app/settings_test.py
@@ -60,3 +60,5 @@ REGISTRATION_ADMINS = (
     ('admin1', 'registration_admin1@mail.server.com'),
     ('admin2', 'registration_admin2@mail.server.com'),
 )
+
+ACCOUNT_ACTIVATION_DAYS = 7
--- a/registration/models.py
+++ b/registration/models.py
@@ -187,8 +187,12 @@ class RegistrationManager(models.Manager):
             registration_profile = self.create_profile(
                 new_user, **profile_info)
 
-        if send_email:
-            registration_profile.send_activation_email(site, request)
+            # send email only if desired and transaction succeeds
+            if send_email:
+                transaction.on_commit(
+                    lambda: registration_profile.send_activation_email(
+                        site, request)
+                )
 
         return new_user
 
--- a/registration/tests/default_backend.py
+++ b/registration/tests/default_backend.py
@@ -3,7 +3,7 @@ import datetime
 from django.conf import settings
 from django.contrib.auth.models import AnonymousUser
 from django.core import mail
-from django.test import TestCase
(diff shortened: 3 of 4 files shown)


test_app/settings_test.py
```diff
@@ -60,3 +60,5 @@
     ('admin1', 'registration_admin1@mail.server.com'),
     ('admin2', 'registration_admin2@mail.server.com'),
 )
+
+ACCOUNT_ACTIVATION_DAYS = 7
```
joshblum: this shouldn't be necessary since we use `override_settings` on the test cases. why is this change necessary? are the overrides no longer working?

joshblum:
thanks for the PR! this change looks good, will accept once the testing issue is resolved

test_app/settings_test.py
adibo: The tests did not pass on a fresh fork from master, before I made any changes (fresh clone). 
Adding the `ACCOUNT_ACTIVATION_DAYS` to `settings_test.py` helped. 
I did not investigate it further, though. I assumed that it was  accidentally removed.
Let's look into it!
Could you check the results of `invoke test` on master?

joshblum: seems to be a weird interaction between pytest/pytest-django and our setup in recent versions.

adibo: If I run tests without `ACCOUNT_ACTIVATION_DAYS` in the `settings_test.py` 5 tests from `default_backend.DefaultBackendViewTests` fail (only from that file). 
If I then apply `@override_settings(ACCOUNTS_ACTIVATION_DAYS=7)` to those (failing) tests, they pass.
This result suggests that there is some problem with `@override_setting` when used jointly at class and method/test level.

Without going any deeper, one way to resolve the issue is to add the missing line to those tests. We then can safely remove `ACCOUNT_ACTIVATION_DAYS` from the `settings_test.py`.
Any preference or suggestion?

joshblum: i'm ok with leaving the setting in `settings_test.py`, thanks for the patch!

adibo: Great! Thanks!

PR #362, opened 2019-04-12
Updated to use SHA256 instead of SHA1

jialutu:
Updated activation_key to use SHA256 from SHA1.

While SHA512 is faster for 64bit systems, I've ultimately decided on SHA256 due to it creating 64 characters as opposed to SHA512 creating 128 characters.


jialutu:
Sorry for the delay. The tests have been added. I've also corrected the test_activation_key_backwards_compatibility tests that I've changed in my previous commits.

registration/tests/models.py
```diff
@@ -650,6 +653,38 @@ def old_method(self, save=True):
         profile = self.registration_profile.objects.get(user=new_user)
         self.assertTrue(profile.activated)
 
+    def test_activation_key_backwards_compatibility_sha1(self):
+        """
+        Make sure that users created witht the old create_new_activation_key method can still be
```
joshblum:
```suggestion
        Make sure that users created with the old create_new_activation_key method can still be
```

joshblum:
Thanks!

JElchison:
Thanks for a great plugin!  I've used this for years in production with very little maintenance required.

Is there a suggested upgrade path from v2.6 to v2.7?  Inclusion of this PR breaks my website with MySQL database error:
```
DataError at /accounts/register/
(1406, "Data too long for column 'activation_key' at row 1")
```
I know that `activation_key` `max_length` increased from 40 to 64.  It seems like a Django migration should have taken care of this for me, and perhaps I've somehow missed that.

Any tips?  Thanks!

joshblum:
@JElchison thank you! Have you run the migration in https://github.com/jialutu/django-registration/blob/master/registration/migrations/0005_activation_key_sha256.py?

JElchison:
@joshblum Firstly, thanks for your speedy reply. And sorry for my delayed response. I had a medical emergency last week, and am just getting back on my feet.

Secondly, my apologies for the noise. I identified a mistake in my control logic that runs Django migrations. The mistake is completely mine.

Best wishes, and thanks (again) for a great plugin!

joshblum:
@JElchison no problem, glad you're doing better

PR #367, opened 2019-08-10
save user and registration profile in one transaction

pacahon:
Default registration form based on `django.contrib.auth.forms.UserCreationForm` which saves active user to db. This PR prevents it and moves user saving to `registration.models.RegistrationManager.create_inactive_user`.

It is up to the developer to maintain this behavior in case of a custom registering form. Especially when custom form omit the save method since the default `UserModel().objects.create_user` implementation also saves user to db.

### DIFF
--- a/registration/backends/default/views.py
+++ b/registration/backends/default/views.py
@@ -88,7 +88,7 @@ class RegistrationView(BaseRegistrationView):
         site = get_current_site(self.request)
 
         if hasattr(form, 'save'):
-            new_user_instance = form.save()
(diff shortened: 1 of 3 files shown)


pacahon:
I don't understand what travis wants from me. 

```
ERROR: /home/travis/build/macropin/django-registration/registration/tests/default_backend.py Imports are incorrectly sorted.
```
Locally isort doesn't complain about sort order anymore. I'll try to recreate PR without push force.

PR #368, opened 2019-08-10
save user and registration profile in one transaction

pacahon:
Default registration form based on django.contrib.auth.forms.UserCreationForm which saves active user to db. This PR prevents it and moves user saving to registration.models.RegistrationManager.create_inactive_user.

It is up to the developer to maintain this behavior in case of a custom registering form. Especially when custom form omit the save method since the default UserModel().objects.create_user implementation also saves user to db.

### DIFF
--- a/registration/backends/default/views.py
+++ b/registration/backends/default/views.py
@@ -88,7 +88,7 @@ class RegistrationView(BaseRegistrationView):
         site = get_current_site(self.request)
 
         if hasattr(form, 'save'):
-            new_user_instance = form.save()
+            new_user_instance = form.save(commit=False)
         else:
             new_user_instance = (UserModel().objects
                                  .create_user(**form.cleaned_data))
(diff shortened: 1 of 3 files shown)


pacahon:
I don't understand what's wrong with travis and how to debug it. Locally isort doesn't complain about anything in the default_backend.py. `make lint` also works fine.

joshblum:
thanks!

pacahon:
@joshblum travis build is still failing. Is it possible to output some debug info behind `make lint` command? Then I could fix an error. Locally I tested imports order with the same isort version (4.3.21)

joshblum:
adding the `--verbose` flag might help:   
```
-vb, --verbose        Shows verbose output, such as when files are skipped
                        or when a check is successful.
```

pacahon:
I'm still confused.
```
SUCCESS: /Users/pacahon/PycharmProjects/django-registration/registration/tests/default_backend.py Everything Looks Good!
```

joshblum:
was this on CI?

joshblum:
fixed here: https://github.com/macropin/django-registration/pull/370

PR #372, opened 2019-12-03
Replace Deprecated Dependency in Support of Django 3.0

### LINKED ISSUE #371: Django 3.0 Support
Django 3.0 [removes](https://docs.djangoproject.com/en/3.0/releases/3.0/#removed-private-python-2-compatibility-apis) `django.utils.encoding.python_2_unicode_compatible()`, which is used as a [decorator](https://github.com/macropin/django-registration/blob/master/registration/models.py#L295-L296) on the RegistrationProfile model.

Removing this decorator is required for Django 3.0 support.

clarkmoody:
Fixes #371

### DIFF
--- a/registration/models.py
+++ b/registration/models.py
@@ -5,6 +5,7 @@ import hashlib
 import logging
 import re
 import string
+import sys
 import warnings
 
 from django.apps import apps
@@ -22,8 +23,6 @@ from django.utils.module_loading import import_string
 from django.utils.timezone import now as datetime_now
 from django.utils.translation import ugettext_lazy as _
 
-from six import python_2_unicode_compatible
-
 from .users import UserModel
 from .users import UserModelString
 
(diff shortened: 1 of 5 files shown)


perceptiveminds:
In Django 3, the UserCreationForm contains the error message 

> The two password fields didn’t match

(with a non-ASCII `’` character) 

To check for that message, the ’ character needs to be in tests/forms.py. But then tox / lint  fails with a non-ASCII error for Python 2, even when using `# coding = utf-8`.

Do you know a way to fix the code to pass that test for Python 2?
(It should not matter at all, since Django 3 doesn't even run on Python 2)

macropin:
I don't think this is the approach we should take. I'm thinking conditional import and a noop on the decorator if not required.

joshblum:
closing in favor of https://github.com/macropin/django-registration/pull/373

PR #373, opened 2019-12-18
Django 3.0 Support

### LINKED ISSUE #371: Django 3.0 Support

clarkmoody:
Adds a basic no-op wrapper in place of a deprecated decorator from Django 3.0. Updates the `tox` environments to cover Django 3.0 and Python 3.8.

Fixes #371 

This is an alternate approach to that taken in #372

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ python:
   - 3.5
   - 3.6
   - 3.7
+  - 3.8
 
 install:
     pip install -r test-requirements.txt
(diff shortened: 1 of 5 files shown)


clarkmoody:
Ok. Tests finally passing 😅 

Thanks to @perceptiveminds for the example of how to fix the failing test in `forms.py`.

joshblum:
@clarkmoody thanks!

PR #377, opened 2020-01-04
Moved python_2_unicode_compatible dependency from Django to SIX

adlerosn:
Django 3.0 removed `python_2_unicode_compatible` from `django.utils.encoding`. This causes this django app to break.

What Django did in version 2.2 was [just forwarding the same function defined in `six` package](https://github.com/django/django/blob/stable/2.2.x/django/utils/encoding.py#L21). So, if you replace `django.utils.encoding` by `six` on [models.py#L21](https://github.com/macropin/django-registration/blob/012dbc6de505ed862b66ee507af20b654a8ee4b1/registration/models.py#L21), it'll fix the issue.

You might want to bump the version after accepting this pull request.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ python:
   - 3.5
   - 3.6
   - 3.7
+  - 3.8
 
 install:
     pip install -r test-requirements.txt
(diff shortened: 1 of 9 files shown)


joshblum:
closing in favor of https://github.com/macropin/django-registration/pull/373

PR #379, opened 2020-03-04
docs: Fix simple typo, modifiying -> modifying

timgates42:
There is a small typo in docs/faq.rst.

Should read `modifying` rather than `modifiying`.


dicato:
Thanks @timgates42! 👍

PR #382, opened 2020-05-08
Deprecate ugettext in favour of gettext

vtemian:
Since Django 3, `ugettext` throws a deprecated warning and is going to be replaced with `gettext`.
https://docs.djangoproject.com/en/3.0/_modules/django/utils/translation/.


joshblum:
thanks!

PR #407, opened 2021-02-05
Update cleanupregistration output for more understaing

quroom:
This code gives more details for cleanupregistration.
Without this, we don't know there is cleaned user or not.
We can only know, command was run or not.

### DIFF
--- a/registration/models.py
+++ b/registration/models.py
@@ -274,6 +274,7 @@ class RegistrationManager(models.Manager):
             models.Q(user__is_active=False) | models.Q(user=None),
             activated=False,
         )
+        deleted_count = 0
         for profile in profiles:
             try:
                 if profile.activation_key_expired():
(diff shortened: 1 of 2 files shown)


quroom:
Fix line error.
Let's see.
If there is more problems, I will look into laster.
Have a nice day.

quroom:
Add new line.
By the way, is there any command to check line in win system?
I can only check invoke test and build.
Any lint test command?

registration/models.py
```diff
@@ -274,16 +274,19 @@ def delete_expired_users(self):
...
             try:
                 if profile.activation_key_expired():
                     user = profile.user
                     logger.warning('Deleting expired Registration profile {} and user {}.'.format(profile, user))
                     profile.delete()
                     user.delete()
+                    deleted_count += 1
             except UserModel().DoesNotExist:
                 logger.warning('Deleting expired Registration profile {}'.format(profile))
                 profile.delete()
```
joshblum, line 288: i think you should increment the count here as well since it results in a profile being deleted

quroom: I missed it I will send pr again.

joshblum (approved):

quroom:
Thanks for review.
Have a nice year.

PR #411, opened 2021-04-02
Add resend activation email validation

quroom:
Before send pr perfectly, I want to make sure this logic.
Does it make sense?
If it looks fine, I will add more codes for test and i18n(po mo) and also clean model code.
Give me your opinion.
If there is better way, let me know it.
I will make it more perfect

And don't merge this code.
I just want to make sure only for logic.

### DIFF
--- a/registration/forms.py
+++ b/registration/forms.py
@@ -9,6 +9,9 @@ you're using a custom model.
 """
 from django import forms
 from django.contrib.auth.forms import UserCreationForm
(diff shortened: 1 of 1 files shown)


quroom:
And can I run lint test on window10?

quroom:
Thanks for great package.
But recently I choose to move auth package to django-allauth.
So I close this pr.

Have a nice day!

Issue #413, opened 2021-04-13
Changing default_auto_field for RegistrationProfile


michalnik:
After upgrade to Django 3.2.x this warning started to appear with `django-admin run` command.
```bash
System check identified some issues:

WARNINGS:
registration.RegistrationProfile: (models.W042) Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField'.
        HINT: Configure the DEFAULT_AUTO_FIELD setting or the AppConfig.default_auto_field attribute to point to a subclass of AutoField, e.g. 'django.db.models.BigAutoField'.
```

It is simply solvable with adding `default_auto_field` to model RegistrationProfile with value `django.db.models.AutoField`

KommuSoft:
This should be put on an app config, not the model: it does not take to that account afaik?

PR #414, opened 2021-04-13
fix https://github.com/macropin/django-registration/issues/413

michalnik:
https://docs.djangoproject.com/en/3.2/releases/3.2/


joshblum:
Thanks!

PR #418, opened 2021-08-31
Update django.po

mnp-fuf:
Mistakes / Typoes fixed


joshblum:
Thank you @zapu and @mnp-fuf !

PR #438, opened 2023-07-06
German translation updated

tobiasmboelz:
Especially wrong fuzzy translations replaced, as well as some wording adjusted to match django.contrib.auth’s translation.


joshblum:
Thanks!

PR #431, opened 2023-02-03
add default_auto_field to the app config, fixing #430

KommuSoft:
See discussion on https://stackoverflow.com/q/75340507/67579

PR #440, opened 2023-10-11
Fix verbose name

Ninjaclasher:
This seems to have been introduced in #431. The `verbose_name` is used in the admin site to generate the title for the app. The current name seems inappropriate for this purpose.
image

This PR changes the name back to the original name before #431.
image


joshblum:
thanks!

PR #441, opened 2023-10-23
Skip deleting users with protected/restricted relations

Urth:
Allow the registration profile cleanup to succeed when a user has protected or restricted relations. The cleanup assumes the users that are selected for cleanup have never been activated. But it is possible for an admin to activate a user without using the registration profile. When the user is then deactivated and the profile becomes a valid candidate for cleanup it may have associated records which prevent it from being deleted.


joshblum (commented):
We may as well move the `profile.delete()` call into the `try` block as well, for good measure. Thanks!

